PR #2, opened 2017-11-26
Dynasource changes

samdark:


src/Container.php
```diff
@@ -49,77 +40,106 @@ class Container implements ContainerInterface
...
-        $this->getting[$id] = 1;
+        return $this->objects[$id];
+    }
 
+    /**
+     * Gets the definition by ID
+     * @param $id
+     * @return array|callable|null|object
+     */
+    private function getDefinition($id)
```
samdark: This method is used only once. Not sure it worth extracting it.

samdark:
Can't be merged as is but some changes should be applied. I'll do it.

PR #3, opened 2017-11-26
Re-done to simple container + builder

samdark:


samdark:
Decided not to go this way.

PR #4, opened 2017-12-15
Added fallback for `__class` to `class` for BC

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | BC with old Yii notation

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -154,6 +154,10 @@ class Container implements ContainerInterface
      * ]);
      * ```
      *
+     * Note: you can use `class` token instead of `__class` for the class name specification, allowing
+     * backward compatibility with the old Yii notation. However this feature is deprecated and will be
+     * removed in future versions.
+     *
      * @param string $id
      * @param mixed $definition
      */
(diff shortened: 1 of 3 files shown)


samdark:
I don't think we need it. It is very likely that a property named `class` will be in the class. That's common for patterns like single table inheritance.

klimov-paul:
> It is very likely that a property named class will be in the class.

Such ability remains, see the unit tests.

> Since we have a BC-breaking release, we have a chance not to introduce features deprecated right from the first version.

I do not like repeating old mistake of Yii1 -> Yii2 migration. I would like to keep as much BC as possible at the initial major release.

klimov-paul:
If accepted, such fallback should be removed somewhere around version 1.0.2 or 1.0.3.

schmunk42:
I'd not include it, to me it adds more confusion. 

I also find it much harder to deprecate it in a future version, ie. if you have 3rd party libs who don't have this fixed they'll work for 1.0.0 and 1.0.1 and people will stay on those, because they can't update their apps.

klimov-paul:
Fallback should be considered while intergating with `yiisoft/yii` package. If there will be such way - it should be considered there. This package should not have such fallabck.
See:
https://github.com/yiisoft/yii2/issues/15436

PR #7, opened 2017-12-15
Di Reference definitions

KoJIT2009:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Add support configure definitions through Reference

#### Input
```
$container = new \yii\di\Container();

$container->set(
    'service',
    [
        '__class' => \yiidiexample\objects\TestService::class,
        'model' => new \yii\di\Reference('model')
    ]
);

$container->set(
    'model',
    \yiidiexample\objects\TestModel::class
);

$service = $container->get('service');
```

#### Output
```
class yiidiexample\objects\TestService#5 (1) {
  public $model =>
  class yiidiexample\objects\TestModel#7 (1) {
    public $property =>
    int(10)
  }
}
```

#### TestService
```
class TestService
{
    /**
     * @var \yiidiexample\objects\TestModel
     */
    public $model;
}
```

#### TestModel
```
class TestModel
{
    /**
     * @var int
     */
    public $property = 10;
}
```

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -235,6 +235,8 @@ class Container implements ContainerInterface
 
         $object = $reflection->newInstanceArgs($dependencies);
 
+        $definition = $this->resolveDependencies($definition);
+
         foreach ($definition as $action => $arguments) {
             if (substr($action, -2) === '()') {
                 // method call
--- a/tests/ContainerTest.php
+++ b/tests/ContainerTest.php
@@ -10,6 +10,7 @@ use yii\di\InvalidConfigException;
 use yii\di\NotFoundException;
 use yii\di\Reference;
 use yii\di\tests\code\Car;
+use yii\di\tests\code\ColorPink;
 use yii\di\tests\code\ConstructorTestClass;
 use yii\di\tests\code\EngineInterface;
 use yii\di\tests\code\EngineMarkOne;
@@ -181,12 +182,15 @@ class ContainerTest extends TestCase
     {
         $container = new Container([
             'engine' => EngineMarkOne::class,
+            'color' => ColorPink::class,
             'car' => [
                 '__class' => Car::class,
                 '__construct()' => [new Reference('engine')],
+                'color' => new Reference('color')
             ],
         ]);
         $object = $container->get('car');
         $this->assertInstanceOf(Car::class, $object);
+        $this->assertInstanceOf(ColorPink::class, $object->color);
     }
 }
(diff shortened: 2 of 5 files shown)


samdark:
I think it could solve #7 and in a better way. Would make configs a flatter @klimov-paul?

klimov-paul:
I would rather avoid usage of `Reference`.
My opinion: `Container` should detect class from the typehinting automatically and compose object from the configuration array of DI item id.
E.g. following code should work:

```php
$container->set('car', [
    '__class' => Car::class,
    '__construct()' => [['__class' => Engine::class]],
    '__setColor()' => [['__class' => ColorPink::class]],
]);

$object = $container->get('car');
```

as well as following:

```php
$container->set('default-engine', [
    '__class' => Engine::class
]);

$container->set('pink-color', [
    '__class' => ColorPink::class
]);

$container->set('car', [
    '__class' => Car::class,
    '__construct()' => [['default-engine']],
    '__setColor()' => [['pink-color']],
]);
$object = $container->get('car');
```

klimov-paul:
Take [yii\db\Connection::$schemaCache](https://github.com/yiisoft/yii2/blob/master/framework/db/Connection.php#L214) for example. I should be able to configure cache component for DB connection instance using a DI:

```php
$container->set('db', [
    '__class' => Connection::class,
    'setSchemaCache()' => [['__class' => FileCache::class]],
]);

$db = $container->get('db');
```

klimov-paul:
Although, I can not tell for sure., someone's else opinion should be asked.

samdark:
How should it tell apart regular arrays from configuration arrays? By presence of `__class`?

samdark:
For example, the following case. Property injection:

```php
'color' => [['pink-color']],
```

I can't tell for sure if it's a reference to another component of just an array.

klimov-paul:
This can be done only via `Reflection` mechanism, by analysing method argument's type restrictions.
However, this would cause a significant performance impact.

I can not say for sure if such automatic detection should be present, but is not DI is about magickal pick up of the container item depending on the method constructor type-hinting?
E.g. determing constructor like:

```php
public function __construct(\yii\db\Connection $db) {}
```

and instantiating object via DI, I expect `\yii\db\Connection` instance to be picked up automatically.

Still, I do not know all expected requirements for the DI container since I am not expert on DI usage.

klimov-paul:
I can not claim proposed approach is incorrect, since I do not really know which it should be.

dynasource:
I am positive towards the use of Reference. It is less magic than Reflection and therefore less confusing to use.

samdark:
These additions should be reflected in the readme.

samdark:
Merged. Thanks!

Issue #10, opened 2018-01-08
[Feature Proposal] Add Decorators


prowwid:
Through container, we can set any default values, attach behaviors, event handlers etc. 

But if there is a bootstrap code that used in different container definitions, add behavior or state to individual objects at run-time or if you want to apply several transformations to an object returned from the container, you need to add a lot of additional code with a meaning that won't be clean to understand sometimes and that won't be really handy.

**I suggest adding support for [decorators](https://sourcemaking.com/design_patterns/decorator) in the container.** 

**Specification**

Each decorator has a method `decorate`  with one argument - object that should be decorated. No other methods are required, so the interface looks like:
```php
/**
 * Represents decorator of any objects.
 *
 * @see https://sourcemaking.com/design_patterns/decorator
 */
interface ObjectDecorator {
    public function decorate($object);
}
```

<hr/>

Decorator can be assigned to a class by its name or identifier in the container, using  method `addDecorator` of the container:
```
$container->addDecorator(Calculator::class, EngineeringDecorator::class);
$container->addDecorator(Calculator::class, ScienceDecorator::class);
```
or through configuration:
```php
'container' => [
    'decorators' => [
        Calculator::class =>  [
            EngineeringDecorator::class,
            ScienceDecorator::class,
        ]
    ]
]
```

<hr/>

Each decorator is a singleton and should be created only one time when a target object is requested from the container.

samdark:
Why should we have decorators separate from other dependencies?

prowwid:
Hello @samdark ,

Sorry, didn't get what you are asking. Could you please explain with more details?

I would expand the idea with better examples and description a bit later today.

With best regards,
Dmitry

samdark:
I don't see how decorator itself matches the concept of DI container i.e. you can store decorator into container but I see no reason why this case should be special. Same as storing non-decorated object, you're setting an object using a named key and it doesn't matter which type of object it is.

prowwid:
@samdark thank you for cleaning things up

**First of all, let's talk about Decorators and how they can be used in Yii.** 

A typical Decorator can look like:
```php
class Book {
    private $author;
    private $title;

    public function __construct($title, $author) {
        $this->author = $author;
        $this->title  = $title;
    }

    public function getAuthor() {
        return $this->author;
    }

    public function getTitle() {
        return $this->title;
    }
    public function getAuthorAndTitle() {
      return $this->getTitle() . ' by ' . $this->getAuthor();
    }
}

[...]

PR #12, opened 2018-01-18
Implemented decorators feature.

prowwid:
Basic implementation of[\[Feature Proposal\] Add Decorators](https://github.com/yiisoft/di/issues/10).

This PR contains a basic implementation of decorators to show how suggested decorators would look like. Examples of decorators present in tests. 

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes


prowwid:
Hello @samdark

Any new thoughts on this one?  
Maybe I need to add something?

samdark:
Technically it is alright but I feel the same could be achieved in a simpler way.

Need opinions of @yiisoft/core-developers and @yiisoft/reviewers.

src/Container.php
```diff
@@ -417,4 +428,58 @@ protected function buildProvider($providerDefinition)
...
+    public function addDecorator($id, $decorator)
+    {
+        if ($this->decoratorHasValidType($decorator)) {
+            $this->decorators[$id][] = $decorator;
+        } else {
+            throw new InvalidConfigException('Decorator should be a callable or implement ' . Decorator::class);
+        }
+    }
+
+    protected function decoratorHasValidType($decorator)
```
SilverFire: How about `isDecorator($object)`?

src/Container.php
```diff
@@ -417,4 +428,58 @@ protected function buildProvider($providerDefinition)
...
+        if (is_callable($decorator)) {
+            $isValid = true;
+        } elseif (is_object($decorator) && $decorator instanceof Decorator) {
+            $isValid = true;
+        } elseif (is_string($decorator) && class_exists($decorator) && (new ReflectionClass($decorator))->implementsInterface(Decorator::class)) {
+            $isValid = true;
+        } else {
+            $isValid = false;
+        }
+        return $isValid;
```
SilverFire: Does this look more readable for you?

```php
if (is_callable($decorator)) {
    return true;
}
if (is_object($decorator) && $decorator instanceof Decorator) {
    return true;
}
if (is_string($decorator) && class_exists($decorator) && (new ReflectionClass($decorator))->implementsInterface(Decorator::class)) {
    return true;
}

return false;
```

src/Container.php
```diff
@@ -417,4 +428,58 @@ protected function buildProvider($providerDefinition)
 
         return $provider;
     }
+
+    public function addDecorator($id, $decorator)
```
SilverFire: `$id` should be renamed to `$groupName` for consistency with `runDecoratorsOnObject()` method

samdark: phpdoc should be added as well...

src/Container.php
```diff
@@ -56,6 +57,14 @@ class Container implements ContainerInterface
      * delayed to register till their services would be requested
      */
     private $delayedProviders;
+    /**
+     * @var contracts\Decorator[]|array
```
samdark: Description should be added.

src/Container.php
samdark: phpdoc should be added.

src/Container.php
```diff
@@ -417,4 +428,58 @@ protected function buildProvider($providerDefinition)
...
+            $isValid = true;
+        } elseif (is_string($decorator) && class_exists($decorator) && (new ReflectionClass($decorator))->implementsInterface(Decorator::class)) {
+            $isValid = true;
+        } else {
+            $isValid = false;
+        }
+        return $isValid;
+    }
+
+    protected function runDecoratorsOnObject($decoratorsGroupName, $object)
```
samdark: phpdoc should be added.

src/Container.php
```diff
@@ -417,4 +428,58 @@ protected function buildProvider($providerDefinition)
...
+    {
+        if (!isset($this->decorators[$decoratorsGroupName]) || empty($this->decorators[$decoratorsGroupName])) {
```
samdark: Why do we need `!isset` check here?

src/Container.php
```diff
@@ -417,4 +428,58 @@ protected function buildProvider($providerDefinition)
...
+                    '__class' => $decorator,
+                ]);
+
+                $this->decorators[$index] = $decorator;
+            }
+            $decorator->decorate($object);
+        }
+    }
+
+    protected function getInjector(): Injector
```
samdark: phpdoc should be added.

src/contracts/Decorator.php
```diff
@@ -0,0 +1,21 @@
...
+ * Represents decorator of objects fetched from the container.
+ *
+ * @see https://sourcemaking.com/design_patterns/decorator
+ *
+ * @author Dmitry Kolodko <prowwid@gmail.com>
+ * @since 1.0
+ */
+interface Decorator
+{
+    public function decorate($object): void;
```
samdark: phpdoc should be added.

prowwid:
Hello @samdark and @SilverFire 

it was a draft realization for [\[Feature Proposal\] Add Decorators](https://github.com/yiisoft/di/issues/10) to reduce the talk on the issue so I basically didn't want to spend more time than needed in case we won't agree merging the PR.

I would add more docs later.

With best regards,
Dmitry

samdark:
> I would add more docs later

prowwid:
Hello @samdark and @SilverFire 

I've made changes in code based on your requests and added docs. Please review.

With best regards,
Dmitry

samdark:
Made some adjustments.

SilverFire:
Wait wait, we are definitely doing something wrong. The thing we've got makes possible to execute something on object right after its creation. Let's refer to the [Decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) article, and read the following:

> Decorator objects implement the interface of the extended (decorated) object transparently by forwarding all requests to it and perform additional functionality before/after forwarding a request.

So current implementation is not a decorator. It should be called `post-configurator` instead, and it's definitely not what we need to do. This feature is absolutely useless, if the decorated object is immutable and does not have neither setters nor public properties.

There are a lot examples on different programming languages on Wikipedia, but all the decoration examples are about wrapping objects to provide additional features, not about changing decorated object internals.

samdark:
Right. It was same concern on my part in https://github.com/yiisoft/di/issues/10#issuecomment-357382054 and I somehow missed when it went wrong :(

samdark:
Anyway, while implementation is good and idea to have decorators is fine, I don't think we'll merge the pull request.

SilverFire:
What we need is an API to build the following imperative construction in a declarative way:

```php
$car = new RemoteControlledCar(new ElectroCar(new Car()));
```

PR #14, opened 2018-01-29
[minor] use root NS specification for opcoded functions

kalessil:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -124,18 +124,18 @@ class Container implements ContainerInterface
 
         $definition = $this->definitions[$id];
 
-        if (is_string($definition)) {
+        if (\is_string($definition)) {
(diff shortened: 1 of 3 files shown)


rob006:
I prefer to use imports for this - less noise in the code, and result should be the same.

```php
// ...
use funtion count;
```

SilverFire:
Agreed with @rob006

kalessil:
@rob006 , @SilverFire : got it. 

Currently, Php Inspections (EA Extended) doesn't support auto-importing for this case so I have to close the PR.

Issue #23, opened 2018-03-06
Add `Container::create()` method


klimov-paul:
At the present state `Container` is almost useless in Yii Framework scope. It is unable to replace old `Container` from Yii 2.0.x.
In particular there is no way to create an object from array definition taking into account possible singleton definitions or predefined configurations.

Previously this task was implemented inside `Container::get()` method, which accepted constructor arguments and object config as additional parameters:

```php
$foo = $container->get(Foo::class, ['constructor argument'], ['someField' => 'some-value']);
```

Accepting PSR for the container makes impossible to use `get()` for this purpose.

I suggest a separated method `create()` (or `createObject()`) should be introduced to incapsulate former object creation logic.
Method signature:

```php
public function create(array $definition) : object
```

This method should check for registered definitions and use `Container::build()` to create final object.

samdark:
Seems to be the same as https://github.com/yiisoft/di/issues/22.

samdark:
Introducing `create()` makes you use container explicitly which defeats its purpose turning it into service locator...

klimov-paul:
I am not sure I understand.
For me in current state `Container` is useless. I am unable to build object from its definition, because this code is hidden inside protected method `Container::build()`.
If I have an object deifnition like:

```php
[
    '__class' => Foo::class,
    'name' => 'something',
]
```

how I can create an object from it?

At the present state the only option I see is setting this definition with some unique name to `Container` and get it from it:

```php
$definition = [
    '__class' => Foo::class,
    'name' => 'something',
];

$id = uniqid();
$container->set($id, $definition);

$object = $container->get($id);
```

I do not see why shortcut method `create()` makes anything worse in comparison with this code.

samdark:
1. You need to put your definition in container.
2. You type-hint in controller via Injector never calling `get` yourself.

klimov-paul:
So I should always create a method with type-hint and invoke it via `Injector` - this does not feels right.

For example: how behaviors for the component should be initialized with such approach?

https://github.com/yiisoft/yii2/blob/master/framework/base/Component.php#L749-L751

klimov-paul:
How we can define default settings for the widgets to be applied during rendering?
For example to attach `yii\jquery\ActiveFormClientScriptBehavior` to every `ActiveForm` instance?

klimov-paul:
How I can define and use 2 separated instances of  `yii\db\Connection` with suhc approach?
[...]

PR #26, opened 2018-03-18
[WIP] Added factory functionality, see #25

hiqsol:
Not finished. Just to see how it looks like.
Will finish (add tests and possibly better naming) after discussion.

Actually changes are minimal but are shown huge because of renamed file.
Tried redo in two commits - doesn't help. Don't know how to help with it.

Highlights:

- **AbstractContainer**
  - `protected build($id, $config)`
  - `protected buildFromDefinition($definition)`
  - **Factory**
    - `public create($id, $config)` - new instance every time
  - **Container**
    - `private $instances`
    - `public get($id)` - same instance every time

No `sameEveryTime` flag was needed.
Factory has `create` but not `get` and Container vice versa.

### DIFF
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -1,0 +1,413 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\di;
+
+use Psr\Container\ContainerInterface;
+use ReflectionClass;
+use SplObjectStorage;
+use yii\di\contracts\DeferredServiceProviderInterface;
+use yii\di\contracts\ServiceProviderInterface;
+
+/**
+ * Container implements a [dependency injection](http://en.wikipedia.org/wiki/Dependency_injection) container.
+ *
+ * @author Alexander Makarov <sam@rmcreative.ru>
+ * @since 1.0
+ */
+abstract class AbstractContainer implements ContainerInterface
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $parent;
+    /**
+     * @var array object definitions indexed by their types
+     */
+    private $definitions;
+    /**
+     * @var ReflectionClass[] cached ReflectionClass objects indexed by class/interface names
+     */
+    private $reflections = [];
+    /**
+     * @var array
+     */
+    private $aliases;
+    /**
+     * @var array cached dependencies indexed by class/interface names. Each class name
+     * is associated with a list of constructor parameter types or default values.
+     */
+    private $dependencies = [];
+    /**
+     * @var array used to collect ids instantiated during build
+     * to detect circular references
+     */
+    private $getting = [];
+    /**
+     * @var contracts\DeferredServiceProviderInterface[]|\SplObjectStorage list of providers
+     * deferred to register till their services would be requested
+     */
+    private $deferredProviders;
+
+    /**
+     * Container constructor.
+     *
+     * @param array $definitions
+     * @param Container|null $parent
+     *
+     * @throws CircularReferenceException
+     * @throws InvalidConfigException
+     * @throws NotFoundException
+     * @throws NotInstantiableException
+     */
+    public function __construct(array $definitions = [], Container $parent = null)
+    {
+        if (isset($definitions['providers'])) {
+            $providers = $definitions['providers'];
+            unset($definitions['providers']);
+        } else {
+            $providers = [];
+        }
+        $this->definitions = $definitions;
+        $this->parent = $parent;
+
+        $this->deferredProviders = new SplObjectStorage();
+        foreach ($providers as $provider) {
+            $this->addProvider($provider);
+        }
+    }
+
+    /**
+     * Creates new instance of the class definition with dependencies resolved
+     * @param array $definition
+     * @return object the newly built instance of the specified class
+     * @throws CircularReferenceException
+     * @throws InvalidConfigException
+     * @throws NotFoundException if there is nothing registered with alias or interface specified
+     * @throws NotInstantiableException
+     */
+    protected function build($id, array $config = [])
+    {
+        if (isset($this->aliases[$id])) {
+            $id = $this->aliases[$id];
+        }
+
+        if (isset($this->getting[$id])) {
+            throw new CircularReferenceException("Circular reference to \"$id\" detected.");
+        }
+        $this->getting[$id] = 1;
+
+        $this->registerProviderIfDeferredFor($id);
+
+        if (!isset($this->definitions[$id])) {
+            if (isset($config['__class'])) {
+                return $this->buildFromDefinition($config);
+            }
+            if ($this->parent !== null) {
+                return $this->parent->build($id, $config);
+            }
+
+            throw new NotFoundException("No definition for \"$id\" found");
+        }
+
+        $definition = $this->definitions[$id];
+
+        if (is_string($definition)) {
+            $definition = ['__class' => $definition];
+        }
+
+        if (is_array($definition) && !isset($definition[0], $definition[1])) {
+            $object = $this->buildFromDefinition($definition);
+        } elseif (is_callable($definition)) {
+            $object = $definition($this, $config);
+        } elseif (is_object($definition)) {
+            $object = $definition;
+        } else {
+            throw new InvalidConfigException('Unexpected object definition type: ' . gettype($definition));
+        }
+
+        unset($this->getting[$id]);
+
+        return $object;
+    }
+
+    /**
+     * Register providers from {@link deferredProviders} if they provide
+     * definition for given identifier.
+     *
+     * @param string $id class or identifier of a service.
+     */
+    protected function registerProviderIfDeferredFor($id): void
+    {
+        $deferredProviders = $this->deferredProviders;
+        if ($deferredProviders->count() === 0) {
+            return;
+        }
+
+        foreach ($deferredProviders as $provider) {
+            if ($provider->hasDefinitionFor($id)) {
+                $provider->register();
+
+                // provider should be removed after registration to not be registered again
+                $deferredProviders->detach($provider);
+            }
+        }
+    }
+
+    /**
+     * Sets a definition to the container. Definition may be defined multiple ways.
+     *
+     * Interface name as string:
+     *
+     * ```php
+     * $container->set('interface_name', EngineInterface::class);
+     * ```
+     *
+     * A closure:
+     *
+     * ```php
+     * $container->set('closure', function($container) {
+     *     return new MyClass($container->get('db'));
+     * });
+     * ```
+     *
+     * A callable array:
+     *
+     * ```php
+     * $container->set('static_call', [MyClass::class, 'create']);
+     * ```
+     *
+     * A definition array:
+     *
+     * ```php
+     * $container->set('full_definition', [
+     *     '__class' => EngineMarkOne::class,
+     *     '__construct()' => [42],
+     *     'argName' => 'value',
+     *     'setX()' => [42],
+     * ]);
+     * ```
+     *
+     * @param string $id
+     * @param mixed $definition
+     */
+    public function set(string $id, $definition): void
+    {
+        if (isset($this->aliases[$id])) {
+            unset($this->aliases[$id]);
+        }
+
+        $this->definitions[$id] = $definition;
+    }
+
+    /**
+     * Sets multiple definitions at once
+     * @param array $config definitions indexed by their ids
+     */
+    public function setMultiple($config): void
+    {
+        foreach ($config as $id => $definition) {
+            $this->set($id, $definition);
+        }
+    }
+
+    /**
+     * Setting an alias so getting an object from container using $id results
+     * in the same object as using $referenceId
+     *
+     * @param string $id
+     * @param string $referenceId
+     */
+    public function setAlias(string $id, string $referenceId): void
+    {
+        $this->aliases[$id] = $referenceId;
+    }
+
+    /**
+     * Returns a value indicating whether the container has the definition of the specified name.
+     * @param string $id class name, interface name or alias name
+     * @return bool whether the container is able to provide instance of id specified.
+     * @see set()
+     */
+    public function has($id): bool
+    {
+        if (isset($this->aliases[$id])) {
+            $id = $this->aliases[$id];
+        }
+
+        return isset($this->definitions[$id]);
+    }
+
+    /**
+     * Creates an instance of the class definition with dependencies resolved
+     * @param array $definition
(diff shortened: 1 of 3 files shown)


src/Container.php
```diff
@@ -7,82 +7,26 @@
...
-            $providers = $definitions['providers'];
-            unset($definitions['providers']);
-        } else {
-            $providers = [];
-        }
-        $this->definitions = $definitions;
-        $this->parent = $parent;
+        /// XXX don't get it
+        /// is it intended to get different instances for same id?
+        $this->instances[$id] = null;
```
samdark: No. For container it's always the same instance.

samdark: The goal of this line is to be react to calling `set()` runtime i.e. it will start returning new instance even if it was obtained before with old definition.

samdark:
Looks good to me. Let's finish it.

src/Container.php
```diff
@@ -107,319 +47,10 @@ public function get($id)
...
-        if (is_array($definition) && !isset($definition[0], $definition[1])) {
-            $object = $this->build($definition);
-        } elseif (is_callable($definition)) {
-            $object = $definition($this);
-        } elseif (is_object($definition)) {
-            $object = $definition;
-        } else {
-            throw new InvalidConfigException('Unexpected object definition type: ' . gettype($definition));
-        }
+        $object = $this->create($id);
```
klimov-paul: How this can function? There is no `create()` method declaration either at `AbstractContainer` not at `Container`.

src/Factory.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Factory is similar to the Container but creates new object every time.
+ *
+ * @author Andrii Vasyliev <sol@hiqdev.com>
+ * @since 1.0
+ */
+class Factory extends AbstractContainer
+{
+    public function create($id, array $config = [])
```
klimov-paul: I don't get it: when I asked for addition of `create()` method to the `Container` at #23, @samdark refused claiming it contradicts PSR. But in case we have an `AbsractContainer` and its 2 descendants, having `create()` method inside of some of them suddenly becomes totally fine!
Does `Factory` class no longer implements PSR interface and thus PSR rules does not apply to it?

src/AbstractContainer.php
```diff
@@ -0,0 +1,413 @@
...
+        $this->getting[$id] = 1;
+
+        $this->registerProviderIfDeferredFor($id);
+
+        if (!isset($this->definitions[$id])) {
+            if (isset($config['__class'])) {
+                return $this->buildFromDefinition($config);
+            }
+            if ($this->parent !== null) {
+                return $this->parent->build($id, $config);
```
klimov-paul: I can not see how it resolves definition handling.
In case `Factory` will use `Container` as parent, it will be unable to use its defintions, this fucntion will fallback to `Conatiner::build()` and always return a singleton or throw an exception.

src/Factory.php
samdark: Yes. Factory is explicitly non-PSR thing. You've got it right.

klimov-paul: Can you clarify in which way you separate 'PSR thing' from 'non-PSR thing'?
As far as I see `Factory` IS a PSR container as it matches `ContainerInterface`, and thus it can be used with `Injector` while resolving dependency injection.
This makes [your words](https://github.com/yiisoft/di/issues/23#issuecomment-370911950) to be standing for it just as well.

It seems you are interpreting PSR-11 as you please depending on particular situation.

samdark: What I was concerned about is [point 1.3 of the PSR](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md#13-recommended-usage):

> Users SHOULD NOT pass a container into an object so that the object can retrieve its own dependencies. This means the container is used as a Service Locator which is a pattern that is generally discouraged.

My concern is that if we'll allow using container as service locator, developer would you use it exactly like that even for singleton services because feels simpler to do that before you start taking care about high coupling.

If we'll look at [meta document](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container-meta.md#4-recommended-usage-container-psr-and-the-service-locator):

> As an exception, factory objects whose only purpose is to create and return new instances may use the service locator pattern. The factory must then implement an interface so that it can itself be replaced by another factory using the same interface.

It's understandable because there's no other way to get instance configured in runtime from a factory and for that use case it's fine. That's why having separate Container (singletons only, for keeping services) and Factory (non-singletons only, for creating specific objects configured in runtime) sounds like a good solution.

src/AbstractContainer.php
hiqsol: No. `Factory` will use its own definitions as far as given ID is among those definitions.
`Factory` has its own definitions and defaults to parent only `if (!isset($this->definitions[$id]))`.

src/Container.php
hiqsol: Yes. This is what I was asking about.
Container will return different object after definition is changed.
I was not sure if it's ok. I haven't found anything about it in PSR.

src/Factory.php
hiqsol: And if `ContainerInterface` seems not suitable for `Factory` it can be fixed easily by removing the interface from `AbstractContainer`.

src/Container.php
hiqsol: Sorry. Didn't test it at all. It has to be `build()`.

I want to get feedback and then do it properly.

src/Factory.php
klimov-paul: > If we'll look at meta document:

This meta document does not allow `Factory` to implement `ConatinerInterface`. The code snippet from the document is following:

```php
interface FactoryInterface // no `ConatinerInterface` !!!
{
    public function newInstance();
}

class ExampleFactory implements FactoryInterface // no `ConatinerInterface` !!!
{
    public function __construct(ContainerInterface $container)
    {
        // has conatiner as a 'dependency'!!!
        $this->container = $container;
    }

    //...
}
```

> And if ContainerInterface seems not suitable for Factory it can be fixed easily by removing the interface from AbstractContainer.

This simply creates a twisted code, when `Factory` matches `ContainerInterface` but does not implement it.

In case `Factory` does not implement `ContainerInterface` as PSR suggests it can not be used for dependency injection in the way dependency object is created per each owner instance. For example:

```php
class Human
{
    public function __constructor(PassportInterface $passport)
    {
        // Requires new `PassportInterface` instance per each new `Human` instance.
        // ...
    }
}
```

So in the end this PR solves nothing.

samdark: Factory should not be used for dependency injection. It should be used to create objects by explicitly calling it with configuration given.

hiqsol: As about the main point I can add that I understand @samdark point.
It is clearly expressed in PSR. The difference is indeed rather small and mostly about naming.

My idea expressed in this PR is:

- leave `Container` pure and strictly following PSR recommendations
- reuse code to implement generic object `Factory` able to substitute Yii 2.0 `Yii::createObject()`

klimov-paul: And so my example with `Human` and `PassportInterface` is simply ignored...
So in the end we will simply pretend the tasks from real world simply do not exist, will not we?

klimov-paul: > reuse code to implement generic object Factory able to substitute Yii 2.0 Yii::createObject()

It can not fully substitute the former `Yii::createObject()` as it was much more sophisticated, allowing dependency injection of both definitions and singletons at the same time along with the factory ability. Following PSR-11 in the strict way will cause its functionality degradation one way or another.

src/AbstractContainer.php
klimov-paul: I see.

src/Factory.php
klimov-paul: Among [DI implementation from the meta document](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container-meta.md#implementors) there is [PHP-DI](http://php-di.org/) for example, which does not afraid to mix `Container` and `Factory`:
https://github.com/PHP-DI/PHP-DI/blob/master/src/Container.php#L38
https://github.com/PHP-DI/PHP-DI/blob/master/src/Container.php#L178

`Laravel` also grands much more flexibility around the DI usage, solving **all** tasks including 'factory-related' ones.

src/Container.php
samdark: Yes. It's OK.

src/Factory.php
samdark: > And so my example with Human and PassportInterface is simply ignored...

Sorry, needed a bit more time to think.

> This simply creates a twisted code, when Factory matches ContainerInterface but does not implement it.

Since it won't have a get() it won't implement the interface.

In the Human-PassportInterface example, I see no contradiction. Each Human, indeed, requires new instance of Passport. That's the job for a factory. The factory itself relies on its own defintions and, when creating an object it does inject its dependencies in the way similar to service container.

> It can not fully substitute the former Yii::createObject() as it was much more sophisticated, allowing dependency injection of both definitions and singletons at the same time along with the factory ability. Following PSR-11 in the strict way will cause its functionality degradation one way or another.

I'm not sure that it will cause degradation. There's inheritance feature in yiisoft/di that would likely solve "both definitions and singletons" i.e. factory's parent will be service container.

> Among DI implementation from the meta document there is PHP-DI for example, which does not afraid to mix Container and Factory
> Laravel also grands much more flexibility around the DI usage, solving all tasks including 'factory-related' ones.

Yes, indeed. There are two ways solving the problem. Either we can mix service container/factory features into one mighty container and strongly recommend that service container features should be used via automatic reflection-based injection while for factory purposes container could be injected directly or we can split these container/factory and limit the ability to use service container wrong technically. I'm for the second approach.

hiqsol: > It can not fully substitute the former Yii::createObject() as it was much more sophisticated, allowing dependency injection of both definitions and singletons at the same time along with the factory ability

I don't see what is missed. `Factory` allows all 3 ways of use:

- it has configured definitions to create non-singletons
- it creates singletons with parent `Container`
- it has to create new object by given config if given class is not defined in the factory or it's parents. I just realized that this is not implemented now but it can be added for `Factory` and not for `Container` (more differences between them).

Which functions of `createObject` am I missing?

> Following PSR-11 in the strict way will cause its functionality degradation one way or another

This is what I can't agree with. I still think both tasks (PSR strictness and good enough factory) can be implemented one way or another. Implementation proposed by me maybe is not good enough but the question in general doesn't look impossible at all.

klimov-paul: > That's the job for a factory. The factory itself relies on its own defintions and, when creating an object it does inject its dependencies in the way similar to service container.

Which factory?

> I'm not sure that it will cause degradation.

And you are fine with that?
I am not. Your pursuite for strict PSR-11 is suitable only for SOLID frameworks like Symfony, which developers inspire those strict rules. Such approach ends up in creation numeous factory classes over the application, each per each task.

Those factories are redundant in the Yii 2.0.x with it 'poor DI' (as it is claimed). You are going to drop the fundamental Yii feature of using 'universal factory' approach via array class definitions ni order to achieve what - creation of another replica of Symfony framework?

And so all YiiSoft team internal decisions about creation of intermediate '2.1.0'  release without major architecture breaks is just nothing for you, isn't it? This is no longer looks like a team work for me.

klimov-paul: > Which functions of createObject am I missing?

`Factory` can not be used with `Injector` to resolve contructions or method calls.
Besides of `Yii::createObject()`, there is also [Container::invoke()](https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php), which you are forgotten about.

In order to claim PSR-11 support, this repository should provide the container and invoker, which can be used to resolved **all** practical issues, including singletons and factory-defintions. So this combination of 'container + invoker' can be used at the end application without any troubles or doubts.

In case `Factory` implements `ContainerInterface` this may work, but @samdark insists this is forbidden.
Also it is bit weird to use instance of class called `Factory` as an actual container whlie using `Invoker`. This will confuse the developers.

samdark: No, we are not going to drop it. We are going to separate it from container. It's not a big architectural break, `createObject()` in the framework will be the same as it was except maybe `__construct()` syntax.

hiqsol: The only suitable goal for me is to substitute former DI implementation with this one with changes I started in this PR.
It seems to me it's possible. We'll try and see :) If it will not go or not be good enough I'll be first to throw the whole attempt away ;)
At the moment I don't see any real stoppers.

`invoke()` can be added - no problem.
As for `ContainerInterface` in `Factory` I think it should be left there.

> Also it is bit weird to use instance of class called Factory as an actual container whlie using Invoker. This will confuse the developers.
Ok it's weird. Those who want be pure and clean will use `Container`.
Those who want weird and new object every time will use `Factory`.

hiqsol: Let a thousand flowers bloom ;)

samdark: `invoke()` was moved to Injector: https://github.com/yiisoft/di/blob/master/src/Injector.php#L58 and yes, you're right that Factory have to impement PSR-11 in order to be used with it... and that's, indeed, a bit confusing.

samdark: If @hiqsol attempt to put it all together will fail, we'll move back to our previous all-in-one container style.

klimov-paul: > As for ContainerInterface in Factory I think it should be left there.

If it is fine to left it here, then why create a separated `Factory` class at the first place?
What it wrong with simple extention of the `Container` public interface I have asked at #23?
Why 2 different DI container should exist? They will only confuse the developers.

Why we can not create a simple solution instead of twisted one?

samdark: The only reason is that technically simpler variant is easier to use wrong (referencing container itself to obtain services from it) and no matter how well we'll document that it's bad to do it, there will be developer who will use it that way. In case of Container/Factory instead of documenting the right way of using container we're forbidding wrong use technically.

klimov-paul: And so, we create an obstacle for ourselves at `Container` to heroically overcome it later with `Factory`!

This is madness... The Symfony-type madness, to be precise. What is the point of creating restriction, which has a bypass solution right near it, for the developer? This is pointless attempt of re-creation Java, which its code restrictions policy, using PHP.

The developer, which expects PSR container from this repository will use `Container` according to the PSR-11 definition anyway. The developer, which want to use `Container` as a factory will end up with `Factory` anyway. Creating 2-level container hierarchy solves nothing.
Besides, in which way developer will setup definitions and singletons with such 2-level structure?
Which instance `Yii::$container` should hold then `Container` or `Factory`?

hiqsol: I've posted example config in #25

There should not be `Yii::$container` in the end. The `Container` should be used in transitional period.

samdark: Yes. In 2.2 or later we should drop `Yii::$container`.

samdark: > The developer, which want to use Container as a factory will end up with Factory anyway. Creating 2-level container hierarchy solves nothing.

It will solve that service `Container` won't be used as service locator.

PR #36, opened 2018-03-25
Dropped aliases in favour of `Reference`, closes #27

### LINKED ISSUE #27: Drop aliases in favour of Reference
Will allow define aliases in config and it seems to me that implementation might get even simpler.

```php
$container = new Container([
    'engine_one' => \yii\di\Reference::to(EngineInterface::class),
];
```

hiqsol:
| Q             | A
| ------------- | ---
| New feature?  | yes
| Breaks BC?    | yes - removed `Container::setAlias()`
| Tests pass?   | yes
| Fixed issues  | #27 

I've implemented recursive dereferencing with detection of circular references.

Another options like optimizing with dereferencing in `set` or complete prohibition of reference to reference are not easily achievable because of the following scenario:

```php
$container->set('engine-1', EngineMarkOne::class);
$container->set('engine-2', Reference::to('engine-1');
$container->set('engine-1', Reference::to(EngineMarkTwo::class));
```

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -38,10 +38,6 @@ class Container implements ContainerInterface
      */
     private $reflections = [];
     /**
-     * @var array
-     */
-    private $aliases;
-    /**
      * @var array cached dependencies indexed by class/interface names. Each class name
      * is associated with a list of constructor parameter types or default values.
      */
@@ -51,6 +47,11 @@ class Container implements ContainerInterface
      * to detect circular references
      */
     private $getting = [];
+    /**
+     * @var array used to collect ids during dereferencing
+     * to detect circular references
+     */
+    private $dereferencing = [];
     /**
      * @var contracts\DeferredServiceProviderInterface[]|\SplObjectStorage list of providers
      * deferred to register till their services would be requested
@@ -99,9 +100,7 @@ class Container implements ContainerInterface
      */
     public function get($id)
     {
-        if (isset($this->aliases[$id])) {
-            $id = $this->aliases[$id];
-        }
+        $id = $this->dereference($id);
 
         if (isset($this->instances[$id])) {
             return $this->instances[$id];
@@ -207,11 +206,7 @@ class Container implements ContainerInterface
      */
     public function set(string $id, $definition): void
     {
-        $this->instances[$id] = null;
-
-        if (isset($this->aliases[$id])) {
-            unset($this->aliases[$id]);
-        }
+        unset($this->instances[$id]);
 
         $this->definitions[$id] = $definition;
     }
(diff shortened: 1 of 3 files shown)


src/Container.php
```diff
@@ -247,13 +230,30 @@ public function setAlias(string $id, string $referenceId): void
...
 
         return isset($this->definitions[$id]);
     }
 
+    /**
+     * Follows references recursively to find deepest ID.
+     *
+     * @param string $id
+     */
+    public function dereference($id)
```
samdark: Should it be public? I don't think so... private?

alexraputa: How about this structure of the method?

```php
public function dereference($id)
{
    if (!isset($this->definitions[$id]) && !$this->definitions[$id] instanceof Reference) {
        return $id;
    }

    if (isset($this->dereferencing[$id])) {
        throw new CircularReferenceException("Circular reference to \"$id\" detected.");
    }

    $this->dereferencing[$id] = 1;
    $id = $this->dereference($this->definitions[$id]->getId());
    unset($this->dereferencing[$id]);
}
```

hiqsol: yes, I prefer less nesting too

hiqsol:
I've implemented suggested changes.

src/Container.php
```diff
@@ -247,13 +230,33 @@ public function setAlias(string $id, string $referenceId): void
...
-        if (isset($this->aliases[$id])) {
-            $id = $this->aliases[$id];
-        }
+        $id = $this->dereference($id);
 
         return isset($this->definitions[$id]);
     }
 
+    /**
+     * Follows references recursively to find deepest ID.
```
SilverFire: find **the** deepest

src/Container.php
```diff
@@ -247,13 +230,33 @@ public function setAlias(string $id, string $referenceId): void
...
+     *
+     * @param string $id
+     */
```
SilverFire: `@return` is missing

src/Container.php
```diff
@@ -247,13 +230,33 @@ public function setAlias(string $id, string $referenceId): void
...
+    private function dereference($id)
```
SilverFire: Typehint please

hiqsol: As for me it's an open question for the moment.
`has()` and `get()` do not typehint `string` for `$id`.
`Reference::_constructor()` and `getId()` too!

So I think it's better to decide first if `$id` is going to be string forever or it may change later - remember of https://github.com/container-interop/container-interop/issues/6#issuecomment-33132151
And typehint or not everywhere accordingly.

PR #37, opened 2018-04-09
Added `Factory`, see #25

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Related issues | #25, #26  

Actually changes are minimal but are shown huge because of renamed file. Done in three commits to allow explore diffs.

Highlights:

- **AbstractContainer**
  - `protected build($id, $config)`
  - `protected buildFromDefinition($definition)`
  - **Factory**
    - `public create($id, $config)` - new instance every time
  - **Container** implements PSR11
    - `private $instances`
    - `public get($id)` - same instance every time

### DIFF
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -7,7 +7,6 @@
 
 namespace yii\di;
 
-use Psr\Container\ContainerInterface;
 use ReflectionClass;
 use SplObjectStorage;
 use yii\di\contracts\DeferredServiceProviderInterface;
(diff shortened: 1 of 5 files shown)


SilverFire:
`Yii::$di->get()` and `Yii::createObject()` methods with old container implementation work as factory. This pull request does not restore the original behavior, but make is theoretically possible. How are we going to keep a sort of behavioral BC with the existing user-side code?

hiqsol:
Please refine what behavior can not be reproduced?

PR #39, opened 2018-07-06
Fixed to use `mb_substr` instead of substr

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
Merged. Thanks!

cebe:
If you compare against non-multibyte chars on the end of the string, substr and mb_substr are the same, while substr is faster: https://3v4l.org/5T2Hk

cebe:
Oh I see https://github.com/yiisoft/di/pull/39 now, nvm

PR #44, opened 2018-07-10
Fixed error in circular reference detection

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Array of currently building items was not cleaned properly in `build()`.

Also enabled `build()` to create object when given only id.

Also refactored `build` into `builWithDefinition` and
`buildWithoutDefinition` - to have single return and single `unset(building)`
and readability of course.
Renamed `buildFromDefinition` to `buildFromConfig` for consistency.


src/AbstractContainer.php
```diff
@@ -108,37 +108,47 @@ public function build($id, array $config = [])
...
-            if ($this->parent !== null) {
-                return $this->parent->build($id, $config);
-            }
+        unset($this->building[$id]);
+
+        return $object;
+    }
 
-            throw new NotFoundException("No definition for \"$id\" found");
+    protected function buildWithoutDefinition($id, array $config = [])
```
SilverFire: PHPDocs are missing

src/AbstractContainer.php
```diff
@@ -108,37 +108,47 @@ public function build($id, array $config = [])
...
+            return $this->parent->build($id, $config);
+        } elseif (class_exists($id)) {
+            $config['__class'] = $id;
+            return $this->buildFromConfig($config);
         }
 
+        throw new NotFoundException("No definition for \"$id\" found");
+    }
+
+    protected function buildWithDefinition($id, array $config = [])
```
SilverFire: PHPDocs are missing

PR #49, opened 2018-08-14
Correct judgment condition

vistart:
$definitions is defined in the parent class and is private.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Correct the conditions for determining whether a definition exists.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -22,7 +22,7 @@ class Factory extends AbstractContainer implements FactoryInterface
      */
     public function get($id)
     {
-        if (!isset($this->definitions[$id]) && $this->parent !== null) {
+        if ($this->getDefinition($id) == null && $this->parent !== null) {
             return $this->parent->get($id);
         }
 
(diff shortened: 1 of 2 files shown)


hiqsol:
Thank you!

You are right - I've added tests for `Factory::get()`.
Please see 58aa6c6.

But I don't agree with use of `getDefinition` instead of `isset($this->definitions[$id])`.
Why do you think it is right?

vistart:
I think the PHP interpreter has problems determining the accessibility of member variables of a class.
If it is private in the parent class and not defined in this class, and the member variable to be accessed is treated as an array, then the exception should also be thrown instead of returning undefined. But at this point PHP returns undefined.

vistart:
According to the definition of `getDefinition()` method of `AbstractContainer`,
`!isset($this->definitions[$id])` and `$this->getDefinition($id) == null` are logically equivalent.

PR #65, opened 2019-01-02
Check $config content when AbstractContainer::configure called

FabrizioCaldarelli:
Related to issue https://github.com/yiisoft/db/issues/42

I think that it should be better check $config content inside ::configure(), instead delegate this to developer calling.

### DIFF
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -384,13 +384,15 @@ abstract class AbstractContainer implements ContainerInterface
      */
     public static function configure($object, iterable $config)
     {
-        foreach ($config as $action => $arguments) {
-            if (substr($action, -2) === '()') {
-                // method call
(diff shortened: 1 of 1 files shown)


machour:
Let's wait to hear back from the team on yiisoft/db#42

src/AbstractContainer.php
```diff
@@ -384,13 +384,15 @@ protected function initObject($object)
...
     public static function configure($object, iterable $config)
     {
-        foreach ($config as $action => $arguments) {
-            if (substr($action, -2) === '()') {
-                // method call
-                \call_user_func_array([$object, substr($action, 0, -2)], $arguments);
-            } else {
-                // property
-                $object->$action = $arguments;
+        if (!empty($config)) {
```
samdark: That doesn't make sense. If `$config` is iterable and is empty, there's no need to check for emptiness. It will just iterate zero times...

PR #70, opened 2019-01-11
Added getInstances() to return all instances in container

FabrizioCaldarelli:
Here we need to get all instances inside container:

https://github.com/yiisoft/yii-console/blob/13913c845e2dceca1cd97e7ac9708415fa1d49d2/src/controllers/CacheController.php#L259

where 

`$components = $this->app->getComponents();`

becomes

`$components = \yii\helpers\Yii::getContainer()->getInstances()`

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -71,4 +71,13 @@ class Container extends AbstractContainer
 
         return isset($this->instances[$id]);
     }
+
+    /**
+     * Returns all instances set in container
+     * @return array list of instance
+     */
+    public function getInstances() : array
+    {
+        return $this->instances;
+    }
 }


kids-return:
yii2-queue used
```php
    /**
     * @return string command id
     * @throws
     */
    protected function getCommandId()
    {
        foreach (Yii::$app->getComponents(false) as $id => $component) {
            if ($component === $this) {
                return Inflector::camel2id($id);
            }
        }
        throw new InvalidConfigException('Queue must be an application component.');
    }
```
https://github.com/yiisoft/yii2-queue/blob/92b7f241fb31de6474a65a0f3af955b2bb357fda/src/cli/Queue.php#L72

hiqsol:
@FabrizioCaldarelli  No, this change will not work.
In Yii2 `$app->getComponents()` returned definitions NOT instances.
Getting instances not enough cause `findCaches` is supposed to find all defined caches. And they can be not yet instantiated.

So I think, `getDefinitions()` needs to be added and used.

hiqsol:
But `getInstances()` can be useful too in other cases.

FabrizioCaldarelli:
What are differences between definitions and instances?

Definitions contains class with $config parameters instead Instances are the relative objects?

hiqsol:
A definition is a definition (e.g. configuration array or closure) and an instance is an object instantiated using given definition.

FabrizioCaldarelli:
> A definition is a definition (e.g. configuration array or closure) and an instance is an object instantiated using given definition.

When an Instance is created from a definition? When I call Yii::getContainer()->set or ->setAll ?

FabrizioCaldarelli:
> > A definition is a definition (e.g. configuration array or closure) and an instance is an object instantiated using given definition.
> 
> When is an Instance created from a definition? When I call Yii::getContainer()->set or ->setAll ?

hiqsol:
When it is got with `get()` or through DI.
This is for performance. There can be tons of definitions in container actually used in different parts of the application or even never used. But definitions get instantiated only when they are required with get or DI.

FabrizioCaldarelli:
What do you think If I launch un for loop to instantiate all definitions before returns from getInstances()?

We'd guarantee that all instances are created.

Issue #75, opened 2019-03-25
ResolveInterface for Reference Resolving


samdark:
See https://forum.yiiframework.com/t/di-reference-resolving-proposing-resolveinterface/125475

samdark:
#76 should solve it.

PR #77, opened 2019-05-01
ResolveInterface for referencing objects implemented

ancpru:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #75 

Declares the interface ResolveInterface with the method get which is called by the container to resolve the reference.

all referencing objects can be resolved by `$theReference->get($theContainerToAsk);`

- `Reference` now implements this interface 
- `ReferencingArray` can be used if a constructor argument or 
  property requires an array containing references  which need to be resolved
- `CallableReference` is a special case. It just contains a callable which is called when container resolves 
  references. This can be used for function calls for constructor arguments or properties which are not valid at the time the container is configured by an configuration array (e.g. resolve aliased path)

### DIFF
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -450,8 +450,8 @@ abstract class AbstractContainer implements ContainerInterface
     {
         foreach ($dependencies as $index => $dependency) {
             if ($dependency instanceof Reference) {
-                if ($dependency->getId() !== null) {
-                    $dependencies[$index] = $this->get($dependency->getId());
+                if ($dependency->isDefined()) {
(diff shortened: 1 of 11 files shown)


machour:
Heads up @ancpru there's a big rework going on at https://github.com/yiisoft/di/pull/76
You may want to check it out and contribute to that branch instead.

ancpru:
> Heads up @ancpru there's a big rework going on at #76
> You may want to check it out and contribute to that branch instead.

Ah, OK. Thanks. Going to rebase to that.

PR #82, opened 2019-06-08
Refactored `Factory` to NOT extend `Container`

hiqsol:
- `Factory` has to implement `ContainerInterface` to be able to
  use `Definition`s that work with `ContainerInterface`.
- temporary solution with `is_string` in `getDefinition()` to prevent resolving loop
- we need MORE tests to understand is such Factory suitable for us


src/Factory.php
```diff
@@ -7,22 +7,94 @@
...
+     */
+    private $definitions = [];
+
+    /**
+     * Factory constructor.
+     *
+     * @param array $definitions
+     * @param ContainerInterface $container
+     * @throws InvalidConfigException
+     * @throws NotInstantiableException
```
samdark: Is it thrown here?

src/Factory.php
```diff
@@ -7,22 +7,94 @@
...
+     */
+    public function __construct(ContainerInterface $container, array $definitions = [])
+    {
+        $this->container = $container;
```
samdark: Field should be declared in a class.

src/Factory.php
```diff
@@ -7,22 +7,94 @@
...
+
+    public function getDefinition($id): Definition
+    {
+        if ($this->has($id)) {
+            return $this->definitions[$id];
+        }
+
+        // XXX out of nowhere solution, without it infinite loop
+        if (\is_string($id)) {
+            return new ArrayDefinition($id);
```
samdark: Can we use simpler `ClassDefinition` here?

samdark: Is it still supporting `Reference::to()`?

hiqsol: > Is it still supporting `Reference::to()`?

we need more tests :)

src/Factory.php
samdark: Is it used?

src/Factory.php
hiqsol: It could be thrown by `ArrayDefinition`

src/Factory.php
hiqsol: > Can we use simpler `ClassDefinition` here?

No. It causes infinite resolving loop.

src/Factory.php
hiqsol: No. It s not used :(
Added test for it.

src/Factory.php
hiqsol: Yeap. We need more test and proper solution for the resolving loop.

hiqsol:
Fixed creating object with dependencies from the parent container.

hiqsol:
Help needed :)

The implementation has the following issues:

- temporary solution with `is_string` in `getDefinition()` to prevent infinite resolving loop
- public `Factory::$container` to allow the following hack:
- `$container = $container->container ?? $container` in `ArrayBuilder::resolveDependencies()`

And what is even more important: there are NOT enough tests!

src/Factory.php
```diff
@@ -7,22 +7,99 @@
...
+
     /**
      * {@inheritdoc}
      */
     public function create($config, array $params = [])
     {
-        $definition = Normalizer::normalize($config);
-
-        return $definition->resolve($this, $params);
+        return Normalizer::normalize($config)->resolve($this, $params);
```
samdark: That means factory works only with container supporting params, right?

hiqsol: Not exactly. It will work with any container. But params will be ignored.
So the factory will create a new object every `create` or `get` call but given params will not be used.

hiqsol: > Not exactly. It will work with any container. But params will be ignored.
> So the factory will create a new object every `create` or `get` call but given params will not be used.

No. I was wrong.

The factory creates configured objects on itself and uses its container only for resolving dependencies. The params will be used only for direct object creating and they are not supposed to be passed into the container.

So the right answer is no. The factory will work with any container.
But I can be wrong :) We need more tests!

hiqsol:
This PR is obsolete by #85 
So I close it.

PR #85, opened 2019-06-22
Separate factory

hiqsol:
Separated factory to own package [yiisoft/factory](https://github.com/yiisoft/factory).
And all the functionality concerning objects creation from definitions moved to the factory package.

So this package got greatly simplified and now contains only:
- containers: default one and composite
- service providers


samdark:
Changes are OK but tests are broken: https://travis-ci.org/yiisoft/di/jobs/549130337#L307

PR #89, opened 2019-09-23
Fix of benchmarks

rugabarbo:
1) Fixed namespaces of classes that were moved to package `factory`
2) Fixed bench crashes caused by exceptions
3) Useless test bench removed
4) Local PHP extensions, especially XDebug, can drastically affect the performance of benchmark subjects, so they are disabled


samdark:
Merged. Thank you!

PR #93, opened 2019-09-24
Closes #88: fall back to autoloader when class is not in container

### LINKED ISSUE #88: Fall back to autoloader when class is not in container
Currently there's a need to declare everything that is repetitive if we need to set a single class without an interface into container:

```php
\App\ConsoleCommand\CreateUser::class => \App\ConsoleCommand\CreateUser::class,
```

In such cases we can try to fall back to class autoloading.

rugabarbo:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #88

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -187,7 +187,7 @@ class Container implements ContainerInterface
      */
     public function has($id): bool
     {
-        return isset($this->definitions[$id]);
+        return isset($this->definitions[$id]) || (is_string($id) && class_exists($id));
     }
 
     /**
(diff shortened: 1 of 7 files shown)


samdark:
Merged. Thank you!

src/Container.php
```diff
@@ -187,7 +187,7 @@ public function setMultiple(array $config): void
      */
     public function has($id): bool
     {
-        return isset($this->definitions[$id]);
+        return isset($this->definitions[$id]) || (is_string($id) && class_exists($id));
```
rob006: Do we need this `is_string()` check here? This method already accepts only string as `$id` and passing for example array will fail at `isset()` level.

rugabarbo: Yes, @rob006, you're right, the check is unnecessary. 
See #96.

PR #95, opened 2019-09-24
Generated missing PHPDocs, Readme corrections

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | comma-separated list of tickets # fixed by the PR, if any

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ While it's usually not a good idea, you can set already instantiated object into
 Additionally, definitions could be added via calling `set()`:
 
 ```php
-/** @var $container \Yiisoft\Di\Container */
+/** @var \Yiisoft\Di\Container $container */
 $container->set($id, Example::class);
 
 $container->set($id, [
@@ -89,7 +89,7 @@ $container->set($id, [
 After container is configured, dependencies could be obtained via `get()`:
 
 ```php
-/** @var $container \Yiisoft\Di\Container */
+/** @var \Yiisoft\Di\Container $container */
 $object = $container->get('interface_name');
 ```
 
@@ -122,7 +122,7 @@ use Yiisoft\Di\CompositeContainer;
 use Yiisoft\Di\Container;
 
 $composite = new CompositeContainer();
-$container = new Container([], [], $composite);
+$container = new Container([], []);
 ```
 
 ## Contextual containers
@@ -232,7 +232,7 @@ To add service provider to the container you need either pass service provider c
 to `addProvider` method of the container:
 
 ```php
-/** @var $container \Yiisoft\Di\Container */
+/** @var \Yiisoft\Di\Container $container */
 $container->addProvider(CarFactoryProvider::class);
 ```
 
@@ -310,7 +310,7 @@ class CarFactoryProvider extends DeferredServiceProvider
     }
 }
 
-/** @var $container \Yiisoft\Di\Container */
+/** @var \Yiisoft\Di\Container $container */
 $container->addProvider(CarFactoryProvider::class);
 
 // returns false as provider wasn't registered
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ use Yiisoft\Di\Container;
 
 $composite = new CompositeContainer();
 $container = new Container([], []);
+$composite->attach($container);
+
 ```
 
 ## Contextual containers
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Usually there is a single container in the whole application so it's often confi
 script such as `index.php` or a configuration file:
 
 ```php
+use Yiisoft\Di\Container;
+
 $container = new Container($config);
 ```
 
@@ -46,7 +48,7 @@ return [
         'argName' => 'value',
         'setX()' => [42],
     ],
-    'closure' => function($container) {
+    'closure' => function(ContainerInterface $container) {
         return new MyClass($container->get('db'));
     },
     'static_call' => [MyFactory::class, 'create'],
@@ -72,6 +74,7 @@ While it's usually not a good idea, you can set already instantiated object into
 Additionally, definitions could be added via calling `set()`:
 
 ```php
+/** @var $container \Yiisoft\Di\Container */
 $container->set($id, Example::class);
 
 $container->set($id, [
@@ -86,6 +89,7 @@ $container->set($id, [
 After container is configured, dependencies could be obtained via `get()`:
 
 ```php
+/** @var $container \Yiisoft\Di\Container */
 $object = $container->get('interface_name');
 ```
 
@@ -98,6 +102,8 @@ Container supports aliases via `Reference` class. It could be useful to have an 
 to retrieve objects both by their interface and named explicitly:
 
 ```php
+use Yiisoft\Di\Container;
+
 $container = new Container([
     EngineInterface::class => EngineMarkOne::class,
     'engine_one' => EngineInterface::class,
@@ -112,6 +118,9 @@ This allows us to combine multiple containers into a composite container which w
 When using this approach, one should only use the composite container.
 
 ```php
+use Yiisoft\Di\CompositeContainer;
+use Yiisoft\Di\Container;
+
 $composite = new CompositeContainer();
 $container = new Container([], [], $composite);
 ```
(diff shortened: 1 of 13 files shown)


src/CompositeContainer.php
```diff
@@ -17,6 +17,9 @@ class CompositeContainer implements ContainerInterface
      */
     private $containers = [];
 
+    /**
+     * @inheritDoc
```
hiqsol: @samdark need to choose uniform inheritdoc syntax to be used everywhere!

tests/Benchmark/ContainerBench.php
```diff
@@ -31,10 +35,12 @@ public function provideDefinitions(): array
         ];
     }
 
-
     /**
      * Load the bulk of the definitions.
      * These all refer to a service that is not yet defined but must be defined in the bench.
+     *
+     * @throws \Yiisoft\Factory\Exceptions\InvalidConfigException
```
hiqsol: Maybe use `use` for usability? :)

tests/Unit/ContainerTest.php
```diff
@@ -295,9 +443,15 @@ public function testGetByClassIndirectly(): void
...
         $this->assertInstanceOf(EngineMarkOne::class, $engine);
-        $this->assertSame($number, $engine->getNumer());
+        $this->assertSame($number, $engine->getNumber());
     }
 
+    /**
+     * @throws \Yiisoft\Factory\Exceptions\CircularReferenceException
+     * @throws \Yiisoft\Factory\Exceptions\InvalidConfigException
+     * @throws \Yiisoft\Factory\Exceptions\NotFoundException
+     * @throws \Yiisoft\Factory\Exceptions\NotInstantiableException
```
hiqsol: I think these functions should not throw anything...

README.md
```diff
@@ -296,6 +310,7 @@ class CarFactoryProvider extends DeferredServiceProvider
     }
 }
 
+/** @var $container \Yiisoft\Di\Container */
```
rob006: I should be either `/* @var $container \Yiisoft\Di\Container */` (vardoc) or `/** @var \Yiisoft\Di\Container $container */` (phpdoc).  `/** @var $container \Yiisoft\Di\Container */` is some weird variation of these two formats.

tests/Unit/ContainerTest.php
xepozz: @hiqsol  Why not?

src/CompositeContainer.php
xepozz: I think we should follow PSR
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#41-making-inheritance-explicit-using-the-inheritdoc-tag

tests/Unit/ContainerTest.php
rob006: These are test actions, they're never used directly. These annotations are useless.

tests/Benchmark/ContainerBench.php
xepozz: But why do `use` if these classes are not used in that file directly?
I always using full classnames in my project.
PHPDocs are not instructions for PHP interpreter thus we should not do `use` imho.

src/CompositeContainer.php
rob006: This PSR is not accepted, and `{@inheritdoc}` is more common in real world. But I would suggest to avoid such phpdoc in the first place - phpdoc is inherited by default, so adding this annotation here does not change anything, just a few redundant lines.

tests/Benchmark/ContainerBench.php
rob006: `use` is more DRY - you don't need to repeat full namespace multiple times, so it is less risk for some inconsistencies.

src/CompositeContainer.php
xepozz: > This PSR is not accepted

It's just a matter of time.
<br>
> `{@inheritdoc}` is more common in real world

I dont't agree with you. 
1. PHPStorm write `inheritDoc` tag by default. 
2. `inheritDoc` it's two words: `inherit` and `doc`. If we want to follow camelCase naming convension we should write upper letter every in following word.
3. PSR-19 
4. [PHPUnit](https://phpunit.readthedocs.io/en/latest/annotations.html) annotations follow camelCase naming convension as well

xepozz: > phpdoc is inherited by default, so adding this annotation here does not change anything, just a few redundant lines.

I agree with you :)
But I was see `@inheritDoc` somewhere and decided to do that.

rob006: > I dont't agree with you.

Then check what other PHP frameworks use: Symfony, Laravel, Zend or Cake uses `{@inheritdoc}`. We already had this discussion in Yii 2 and we switched to `{@inheritdoc}` because it is more common and has better IDE support: https://github.com/yiisoft/yii2/issues/11635

xepozz: Are you seriously?
There are more weightly documents, than Symfony, Laravel or Zend:
1. phpdocumentator https://docs.phpdoc.org/guides/inheritance.html#the-inheritdoc-tag
2. PSR-19
Both are using upper letter.
-------
You are lying me:
1. Zend and Laravel are not using `@inheritDoc` at all.
Proofs: [Zend](https://github.com/zendframework/zendframework/search?q=inheritDoc&unscoped_q=inheritDoc), [Laravel](https://github.com/laravel/laravel/search?q=inheritDoc&unscoped_q=inheritDoc)
2. Zend using upper letter in own issues: https://github.com/zendframework/zendframework/search?q=inheritDoc&type=Issues
3. at the time of writing this document did not exist. 
I sure we need to trust PSR-19/PHPDocumentator rather Symfony.

xepozz: Also i think
> {@inheritdoc} because it is more common and has better IDE support

that's fiction. Why you think so? Give me some proofs please.

rob006: > There are more weightly documents /.../

These are NOT standards.

> Proofs: [Zend](https://github.com/zendframework/zendframework/search?q=inheritDoc&unscoped_q=inheritDoc), [Laravel](https://github.com/laravel/laravel/search?q=inheritDoc&unscoped_q=inheritDoc)

Correct searches: 
1. Zend (it is collection of libraries now - `zendframework/zendframework` is only metapackage): https://github.com/search?q=org%3Azendframework+inheritdoc&type=Code
2. Laravel (`laravel/laravel` is only project starter): https://github.com/laravel/framework/search?q=inheritdoc&type=Code

> 3\. at the time of writing this document did not exist.

PSR-19 was extracted from PSR-5. The specs is pretty old and changed multiple times already.

> that's fiction. Why you think so? Give me some proofs please.

I gave you a link to Yii 2 issue - there is some research here. But in general you should expect that standard which everyone uses has better support than recommendation from in-draft-since-ages-PSR.

xepozz: Ok I will remove these tags.

> These are NOT standards

So let it be. I don't want to waste my time

hiqsol: That's why I started with: "need to choose uniform syntax..." :)
It's not about what others prefer or use. It's just about chosen convention.
So let's wait for @samdark and chose the convention :)
And I wouldn't even notice the @inherit here if it was uniformly used all over this PR which was not.

README.md
```diff
@@ -112,8 +118,11 @@ This allows us to combine multiple containers into a composite container which w
 When using this approach, one should only use the composite container.
 
 ```php
+use Yiisoft\Di\CompositeContainer;
+use Yiisoft\Di\Container;
+
 $composite = new CompositeContainer();
-$container = new Container([], [], $composite);
```
samdark: Why do we need $composite variable then?

src/Contracts/ServiceProviderInterface.php
```diff
@@ -14,11 +14,11 @@
  *
  * Example:
  * ```php
- * class CarProvider implements ServiceProvider
+ * class CarProvider implements ServiceProviderInterface
  * {
  *    protected $container;
  *
- *    public function __construct(ContainerInterface $container)
+ *    public function __construct(Container $container)
```
samdark: It is meant to be interface.

xepozz: But how will you do it  https://github.com/yiisoft/di/blob/060788490c462a520d9d18f64afe2b4c5ebd07ec/src/Contracts/ServiceProviderInterface.php#L35
?

samdark:
Merged. Thanks!

PR #100, opened 2020-01-24
Add composite container to the factory

yiiliveext:
For PR https://github.com/yiisoft/factory/pull/4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^8.4",
+        "league/container": "^3.2",
         "squizlabs/php_codesniffer": "^3.3",
         "phpbench/phpbench": "^0.14.0"
     },
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -48,11 +48,9 @@ class CompositeContainer implements ContainerInterface
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(?ContainerInterface $container): void
+    public function attach(ContainerInterface $container): void
     {
-        if (null !== $container) {
-            array_unshift($this->containers, $container);
-        }
+        array_unshift($this->containers, $container);
     }
 
     /**
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -23,7 +23,7 @@ class CompositeContainer implements ContainerInterface
     {
         foreach ($this->containers as $container) {
             try {
-                if (!empty($parameters)) {
+                if ($parameters !== []) {
                     return $container->get($id, $parameters);
                 }
                 return $container->get($id);
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -27,8 +27,6 @@ class CompositeContainer implements ContainerInterface
                     return $container->get($id, $parameters);
                 }
                 return $container->get($id);
-            } catch (NotFoundExceptionInterface $e) {
-                // ignore
             } catch (ContainerExceptionInterface $e) {
                 // ignore
             }
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -52,7 +52,7 @@ class CompositeContainer implements ContainerInterface
      */
     public function attach(?ContainerInterface $container): void
     {
-        if (isset($container)) {
+        if (null !== $container) {
             array_unshift($this->containers, $container);
         }
     }
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -4,6 +4,7 @@ namespace Yiisoft\Di;
 
 use Psr\Container\ContainerInterface;
 use Psr\Container\NotFoundExceptionInterface;
+use Psr\Container\ContainerExceptionInterface;
 use Yiisoft\Factory\Exceptions\NotFoundException;
 
 /**
@@ -18,12 +19,17 @@ class CompositeContainer implements ContainerInterface
      */
     private $containers = [];
 
-    public function get($id)
+    public function get($id, array $parameters = [])
     {
         foreach ($this->containers as $container) {
             try {
+                if (!empty($parameters)) {
+                    return $container->get($id, $parameters);
+                }
                 return $container->get($id);
             } catch (NotFoundExceptionInterface $e) {
+                // ignore
+            } catch (ContainerExceptionInterface $e) {
                 // ignore
             }
         }
@@ -44,9 +50,11 @@ class CompositeContainer implements ContainerInterface
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(?ContainerInterface $container): void
     {
-        array_unshift($this->containers, $container);
+        if (isset($container)) {
+            array_unshift($this->containers, $container);
+        }
     }
 
     /**
--- a/tests/Unit/AbstractPsrContainerTest.php
+++ b/tests/Unit/AbstractPsrContainerTest.php
@@ -1,0 +1,190 @@
+<?php
+
+namespace Yiisoft\Di\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Psr\Container\ContainerInterface;
+use Psr\Container\NotFoundExceptionInterface;
+use Psr\Container\ContainerExceptionInterface;
+use Yiisoft\Di\Tests\Support\A;
+use Yiisoft\Di\Tests\Support\B;
+use Yiisoft\Di\Tests\Support\Car;
+use Yiisoft\Di\Tests\Support\CarFactory;
+use Yiisoft\Di\Tests\Support\ColorPink;
+use Yiisoft\Di\Tests\Support\ConstructorTestClass;
+use Yiisoft\Di\Tests\Support\Cycle\Chicken;
+use Yiisoft\Di\Tests\Support\Cycle\Egg;
+use Yiisoft\Di\Tests\Support\EngineInterface;
+use Yiisoft\Di\Tests\Support\EngineMarkOne;
+use Yiisoft\Di\Tests\Support\EngineMarkTwo;
+use Yiisoft\Di\Tests\Support\InvokeableCarFactory;
+use Yiisoft\Di\Tests\Support\MethodTestClass;
+use Yiisoft\Di\Tests\Support\PropertyTestClass;
+use Yiisoft\Di\Tests\Support\TreeItem;
+
+/**
+ * General tests for PSR-11 container.
+ * To be extended for specific containers.
+ */
+abstract class AbstractPsrContainerTest extends TestCase
+{
+    abstract public function createContainer(iterable $definitions = []): ContainerInterface;
+
+    abstract public function setupContainer(ContainerInterface $container, iterable $definitions = []): ContainerInterface;
+
+//  !!!Failed for the League Container. Why?
+//    public function testSettingScalars(): void
+//    {
+//        $this->expectException(InvalidConfigException::class);
+//        $container = $this->createContainer([
+//            'scalar' => 123,
+//        ]);
+//
+//        $container->get('scalar');
+//    }
+
+//  !!!Error for the League Container. Why?
+//    public function testOptionalCircularClassDependency(): void
+//    {
+//        $container = $this->createContainer([
+//            A::class => A::class,
+//            B::class => B::class,
+//        ]);
+//        $a = $container->get(A::class);
+//        $this->assertInstanceOf(B::class, $a->b);
+//        $this->assertNull($a->b->a);
+//    }
+
+//  !!!Failed for the League Container. Why?
+//    public function testWithoutDefinition(): void
+//    {
+//        $container =  $this->createContainer();
+//
+//        $hasEngine = $container->has(EngineMarkOne::class);
+//        $this->assertTrue($hasEngine);
+//
+//        $engine = $container->get(EngineMarkOne::class);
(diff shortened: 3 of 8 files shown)


src/CompositeContainer.php
```diff
@@ -18,13 +19,18 @@ class CompositeContainer implements ContainerInterface
      */
     private $containers = [];
 
-    public function get($id)
+    public function get($id, array $parameters = [])
```
samdark: Isn't that a violation of PSR-11?

src/CompositeContainer.php
```diff
@@ -44,9 +50,11 @@ public function has($id)
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(?ContainerInterface $container): void
     {
-        array_unshift($this->containers, $container);
+        if (isset($container)) {
```
rustamwin:
```suggestion
        if (null !== $container) {
```

src/CompositeContainer.php
```diff
@@ -44,9 +50,11 @@ public function has($id)
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(?ContainerInterface $container): void
```
xepozz: Why?

src/CompositeContainer.php
```diff
@@ -18,13 +19,18 @@ class CompositeContainer implements ContainerInterface
...
     {
         foreach ($this->containers as $container) {
             try {
+                if (!empty($parameters)) {
+                    return $container->get($id, $parameters);
+                }
                 return $container->get($id);
             } catch (NotFoundExceptionInterface $e) {
                 // ignore
+            } catch (ContainerExceptionInterface $e) {
```
xepozz: Merge with previous

src/CompositeContainer.php
yiiliveext: Because you can receive a few containers from some source and just attach in a loop without checking for null.

src/CompositeContainer.php
yiiliveext: Merged

src/CompositeContainer.php
yiiliveext: PSR-11 is no good idea. Any sufficiently complex PSR-11 container is not interchangeable with another PSR-11 container. I added tests for PSR-11 container, just look at my commit https://github.com/yiisoft/di/pull/100/commits/f08daeaa64b6289ed9b4eb2c07eb69b1709eb85b.
Even simple tests are failed on another PSR-11 League container.

src/CompositeContainer.php
yiiliveext: Or another example
`public function __construct(ContainerInterface $container = null)
{
    $localContainer = new Container();
    $localContainer->set('someLocalService', SomeLocalService::class);
    $compositeContainer = new CompositeContainer();
    $compositeContainer->attach($container);
    $compositeContainer->attach($localContainer);
    $this->container =  $compositeContainer;
}
`

yiiliveext:
Trying to add a test for a get() method of the Container with parameters.
```
public function testGetWithParams(): void
    {
        $container = new Container();
        $one = $container->get(Car::class, [$container->get(EngineMarkOne::class)]);
        $two = $container->get(Car::class, [$container->get(EngineMarkTwo::class)]);
        $this->assertNotSame($one, $two);
        $this->assertInstanceOf(Car::class, $one);
        $this->assertInstanceOf(Car::class, $two);
        $this->assertInstanceOf(EngineMarkOne::class, $one->getEngine());
        $this->assertInstanceOf(EngineMarkTwo::class, $two->getEngine());

    }
```
The test fails. Both $one and $two is the same object. Is it correct or not?

src/CompositeContainer.php
```diff
@@ -18,12 +19,15 @@ class CompositeContainer implements ContainerInterface
...
+    public function get($id, array $parameters = [])
     {
         foreach ($this->containers as $container) {
             try {
+                if (!empty($parameters)) {
+                    return $container->get($id, $parameters);
+                }
                 return $container->get($id);
-            } catch (NotFoundExceptionInterface $e) {
+            } catch (ContainerExceptionInterface $e) {
```
rustamwin:
```suggestion
            } catch (NotFoundExceptionInterface | ContainerExceptionInterface $e) {
```

yiiliveext: NotFoundExceptionInterface extends ContainerExceptionInterface

composer.json
```diff
@@ -28,6 +28,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^8.4",
+        "league/container": "^3.2",
```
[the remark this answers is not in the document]
yiiliveext: See https://github.com/yiisoft/di/commit/f08daeaa64b6289ed9b4eb2c07eb69b1709eb85b

src/CompositeContainer.php
xepozz: This method is public and it is unacceptable to receive a null argument.

yiiliveext: > This method is public and it is unacceptable to receive a null argument.

https://3v4l.org/v4m6U

src/CompositeContainer.php
yiiliveext: Maybe I said it in the wrong way, PSR-11 is not bad, but not good enough. It would be better if it described such things as definitions, service providers, deflectors, delegation, etc.

src/CompositeContainer.php
xepozz: You didn't convince me. What are the advantages of such a decision?

yiiliveext: This is an ability to disable a container from a higher level. Let's imagine that you have some service with the constructor described above. What will be happened when you set the container parameter to null? Before my fix, an exception will be thrown and the service will not be created. After my fix, the service will be created successfully. And then it all depends on how you use the service. If you call any method that does not use external dependencies then it will be successfully executed, otherwise, an exception will be thrown.

xepozz: How often you use `CompositeContainer` in your projects? For example I never used

yiiliveext: > How often you use `CompositeContainer` in your projects? For example I never used

I use something like CompositeContainer with yii2 apps, but I use own Application, Module, Controller, Container classes.

src/CompositeContainer.php
```diff
@@ -18,12 +19,15 @@ class CompositeContainer implements ContainerInterface
      */
     private $containers = [];
 
-    public function get($id)
+    public function get($id, array $parameters = [])
     {
         foreach ($this->containers as $container) {
             try {
+                if (!empty($parameters)) {
```
viktorprogger:
```suggestion
                if ($parameters !== []) {
```

We should use strict comparison since we have a typehint for a parameter. It is faster and cleaner.

src/CompositeContainer.php
viktorprogger: 1. IMHO service mustn't have direct access to a DI container. This leads to big problems.
1. This function definition with `null` accepted is confusing. When I see I can pass `null` then I think it will be processed somehow. But it is just ignored. I think 
    - function should not accept arguments which will be ignored
    - outer function must be aware of what does it pass as an argument

yiiliveext: 1. I agree, am working on a demo-app that demonstrates my point of view on the concept of modules/services through a container.
2. It's not critical. I can leave as it was and just add the check in the factory PR. Perhaps this fix will be added in the future.

yiiliveext: @samdark, what do you think about?

hiqsol: Agree with @viktorprogger: it makes no sense to attach null.

viktorprogger:
I had a quick look into `league/container` to see if we can fix tests. Unfortunately, both `yiisoft/di` and `league/container` are not fully compatible with PSR-11 and completely incompatible with each other:
- `yiisoft/di` defines `get($id, array $params)`
- `league/container` defines `get($id, bool $new)`

yiiliveext:
This PR can be closed, I've created an alternative PR to fix the factory bug (https://github.com/yiisoft/factory/pull/8)

viktorprogger:
There are lots of tests here and some changes made to `CompositeContainer` which make it fully compatible with the `Yiisoft\Di\Container`. I think we can merge it after resolving of all disagreements.

yiiliveext:
> There are lots of tests here and some changes made to `CompositeContainer` which make it fully compatible with the `Yiisoft\Di\Container`. I think we can merge it after resolving of all disagreements.

Oh, sorry, I forgot it.

yiiliveext:
> I'm trying to add a test for the get() method of the Container with parameters.
> 
> ```
> public function testGetWithParams(): void
>     {
>         $container = new Container();
>         $one = $container->get(Car::class, [$container->get(EngineMarkOne::class)]);
>         $two = $container->get(Car::class, [$container->get(EngineMarkTwo::class)]);
>         $this->assertNotSame($one, $two);
>         $this->assertInstanceOf(Car::class, $one);
>         $this->assertInstanceOf(Car::class, $two);
>         $this->assertInstanceOf(EngineMarkOne::class, $one->getEngine());
>         $this->assertInstanceOf(EngineMarkTwo::class, $two->getEngine());
> 
>     }
> ```
> 
> The test fails. Both $one and $two is the same object. Is it correct or not?

What about this?

viktorprogger:
> What about this?

Yes, this is an expected behavior. Container caches everything it has already built. That's why your second `$container->get(Car::class, ...` does not actually build new `Car`, but returns the one it has built earlier. Normally `$container->get()` is never called explicitly in any project, and all dependencies are declaratively defined in a configuration file.

In the case you want to manually `get` different objects on each call you should use `Factory` from `yiisoft/factory`.

yiiliveext:
> Yes, this is an expected behavior. 

I got it. See https://github.com/yiisoft/factory/issues/6

src/CompositeContainer.php
viktorprogger: @samdark Is this conversation resolved? I think we should leave this parameter for now, but it needs to be deleted later as it is considered for the first call of the method only.

tests/Unit/AbstractPsrContainerTest.php
```diff
@@ -0,0 +1,190 @@
...
+abstract class AbstractPsrContainerTest extends TestCase
+{
+    abstract public function createContainer(iterable $definitions = []): ContainerInterface;
+
+    abstract public function setupContainer(ContainerInterface $container, iterable $definitions = []): ContainerInterface;
+
+    //  !!!Failed for the League Container. Why?
+//    public function testSettingScalars(): void
+//    {
+//        $this->expectException(InvalidConfigException::class);
```
viktorprogger: League container can't throw `Yiisoft\Factory\Exceptions\InvalidConfigException`

tests/Unit/LeaguePsrContainerTest.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Test the League PSR-11 Container.
+ */
+class LeaguePsrContainerTest extends AbstractPsrContainerTest
+{
+    public function createContainer(iterable $definitions = []): ContainerInterface
+    {
+        return $this->setupContainer(new Container(), $definitions);
+    }
```
viktorprogger: Try to use this. It is required to resolve dependencies via autoloading.
```php
    public function createContainer(iterable $definitions = []): ContainerInterface
    {
        $container = new Container();
        $container->delegate(new ReflectionContainer());

        return $this->setupContainer($container, $definitions);
    }
```

viktorprogger: The same is true for composite container tests

tests/Unit/AbstractPsrContainerTest.php
```diff
@@ -0,0 +1,190 @@
...
+//        $engine = $container->get(EngineMarkOne::class);
+//        $this->assertInstanceOf(EngineMarkOne::class, $engine);
+//    }
+
+    public function testCircularClassDependencyWithoutDefinition(): void
+    {
+        $container = $this->createContainer();
+        $this->expectException(ContainerExceptionInterface::class);
+        $container->get(Chicken::class);
+    }
```
viktorprogger: League container doesn't track if dependencies are circular. Php will fall whe on memory end.

tests/Unit/AbstractPsrContainerTest.php
```diff
@@ -0,0 +1,190 @@
...
+    //  !!!Failed for the League Container. Why?
+//    public function testTrivialDefinition(): void
+//    {
+//        $container = $this->createContainer([
+//            EngineMarkOne::class => EngineMarkOne::class,
+//        ]);
+//        $one = $container->get(EngineMarkOne::class);
+//        $two = $container->get(EngineMarkOne::class);
+//        $this->assertInstanceOf(EngineMarkOne::class, $one);
+//        $this->assertSame($one, $two);
```
viktorprogger: League container behaves as a factory by default. This will always fail for it.

tests/Unit/AbstractPsrContainerTest.php
```diff
@@ -0,0 +1,190 @@
...
+//        $this->assertInstanceOf(Car::class, $object);
+//    }
+
+    //  !!!Failed for the League Container. Why?
+//    public function testSameInstance(): void
+//    {
+//        $container = $this->createContainer(['engine' => EngineMarkOne::class]);
+//        $one = $container->get('engine');
+//        $two = $container->get('engine');
+//        $this->assertSame($one, $two);
```
viktorprogger: League container behaves as a factory by default. This will always fail for it.

PR #105, opened 2020-02-18
Add CompositeContainer get params, add tests

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -4,7 +4,6 @@ namespace Yiisoft\Di;
 
 use Psr\Container\ContainerInterface;
 use Psr\Container\NotFoundExceptionInterface;
-use Psr\Container\ContainerExceptionInterface;
 use Yiisoft\Factory\Exceptions\NotFoundException;
 
 /**
@@ -27,7 +26,7 @@ class CompositeContainer implements ContainerInterface
                     return $container->get($id, $parameters);
                 }
                 return $container->get($id);
-            } catch (ContainerExceptionInterface $e) {
+            } catch (NotFoundExceptionInterface $e) {
                 // ignore
             }
         }
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -4,6 +4,7 @@ namespace Yiisoft\Di;
 
 use Psr\Container\ContainerInterface;
 use Psr\Container\NotFoundExceptionInterface;
+use Psr\Container\ContainerExceptionInterface;
 use Yiisoft\Factory\Exceptions\NotFoundException;
 
 /**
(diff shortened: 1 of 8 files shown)


xepozz:
Are there any test cases with using the second argument `parameters`?

src/CompositeContainer.php
```diff
@@ -18,12 +19,15 @@ class CompositeContainer implements ContainerInterface
...
+    public function get($id, array $parameters = [])
     {
         foreach ($this->containers as $container) {
             try {
+                if ($parameters !== []) {
+                    return $container->get($id, $parameters);
+                }
                 return $container->get($id);
-            } catch (NotFoundExceptionInterface $e) {
+            } catch (ContainerExceptionInterface $e) {
```
samdark: Should we ignore errors when instantiating existing dependencies?

src/CompositeContainer.php
```diff
@@ -18,12 +19,15 @@ class CompositeContainer implements ContainerInterface
      */
     private $containers = [];
 
-    public function get($id)
+    public function get($id, array $parameters = [])
```
samdark: https://github.com/yiisoft/di/issues/106

tests/Unit/AbstractCompositePsrContainerTest.php
```diff
@@ -0,0 +1,73 @@
...
+use Yiisoft\Di\Tests\Support\EngineMarkOne;
+use Yiisoft\Di\Tests\Support\EngineMarkTwo;
+
+/**
+ * General tests for PSR-11 composite container.
+ * To be extended for specific containers.
+ */
+abstract class AbstractCompositePsrContainerTest extends AbstractPsrContainerTest
+{
+    public function createCompositeContainer(?ContainerInterface $attachedContainer = null): ContainerInterface
```
samdark: Since we're calling `attach()` and `detach()` on it, it should be more concrete.

```suggestion
    public function createCompositeContainer(?ContainerInterface $attachedContainer = null): CompositeContainer
```

src/CompositeContainer.php
yiiliveext: I think we should, why not?

tests/Unit/AbstractCompositePsrContainerTest.php
yiiliveext: This is just a fast way to attach the first container.

src/CompositeContainer.php
samdark: For example, sources are container1 and container2. Both have A dependency. When we're trying to retrieve A from composite container (container1, container2), out intent is clearly to get dependency from container1. It will fail because of incorrect definition, dependency will be obtained from container 2 and we'll never know why.

tests/Unit/AbstractCompositePsrContainerTest.php
yiiliveext: Fixed

src/CompositeContainer.php
samdark: Will handle this one separately.

samdark:
Merged. Thanks!

Issue #107, opened 2020-02-18
CompositeContainer is not working


xepozz:
### What steps will reproduce the problem?
1. Run the following code:
```php
$container = new \Yiisoft\Di\CompositeContainer();
$container1 = new Container([
    'first' => function () {
        return 'first';
    }
]);
$container2 = new Container([
    'second' => function () {
        return  'second';
    },
    'third' => function ($c) {
        return $c->get('first') . $c->get('second') . 'third';
    },
]);

$container->attach($container1);
$container->attach($container2);
var_dump($container->get('first')); // works
var_dump($container->get('second')); // works
var_dump($container->get('third')); // not working :C
```

### What is the expected result?
Each `$container->get()` should return result
```
string(5) "first" string(6) "second" string(16) "firstsecondthird"
```
### What do you get instead?
```
string(5) "first" string(6) "second"
Fatal error: Uncaught Yiisoft\Factory\Exceptions\CircularReferenceException: Circular reference to "first" detected while building: first,third in /app/vendor/yiisoft/di/src/Container.php:105 Stack trace: #0 /app/vendor/yiisoft/di/src/Container.php(80): Yiisoft\Di\Container->build('first', Array) #1 /app/public/index.php(18): Yiisoft\Di\Container->get('first') #2 /app/vendor/yiisoft/factory/src/Definitions/CallableDefinition.php(19): {closure}(Object(Yiisoft\Di\Container), Array) #3 /app/vendor/yiisoft/di/src/Container.php(134): Yiisoft\Factory\Definitions\CallableDefinition->resolve(Object(Yiisoft\Di\Container), Array) #4 /app/vendor/yiisoft/di/src/Container.php(114): Yiisoft\Di\Container->buildInternal('third', Array) #5 /app/vendor/yiisoft/di/src/Container.php(80): Yiisoft\Di\Container->build('third', Array) #6 /app/vendor/yiisoft/di/src/CompositeContainer.php(25): Yiisoft\Di\Container->get('third') #7 /app/public/index.php(25): Yiisoft\Di\CompositeContainer->get('third') #8 {main} thrown in /app/vendor/yiisoft/di/src/Container.php on line 105
```

Any idea how to solve this?

-------

### UPD:
If try to do `$container->get('third')` only, then we get:
```
[...]

PR #109, opened 2020-02-18
Test case for https://github.com/yiisoft/di/issues/107

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | https://github.com/yiisoft/di/issues/107

### DIFF
--- a/tests/Unit/CompositeContainerTest.php
+++ b/tests/Unit/CompositeContainerTest.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Yiisoft\Di\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Yiisoft\Di\CompositeContainer;
+use Yiisoft\Di\Container;
+
+class CompositeContainerTest extends TestCase
+{
+    public function testFallbackContainer(): void
+    {
+        $container = new CompositeContainer();
+        $container->attach(
+            new Container(
+                [
+                    'first' => function () {
+                        return 'first';
+                    },
+                ]
+            )
+        );
+        $container->attach(
+            new Container(
+                [
+                    'second' => function () {
+                        return 'second';
+                    },
+                    'third' => function ($c) {
+                        return $c->get('first') . $c->get('second') . 'third';
+                    },
+                ]
+            )
+        );
+        $this->assertSame('first', $container->get('first'));
+        $this->assertSame('second', $container->get('second'));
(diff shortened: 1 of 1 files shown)


tests/Unit/CompositeContainerTest.php
```diff
@@ -0,0 +1,39 @@
...
+            )
+        );
+        $container->attach(
+            new Container(
+                [
+                    'second' => function () {
+                        return 'second';
+                    },
+                    'third' => function ($c) {
+                        return $c->get('first') . $c->get('second') . 'third';
```
viktorprogger: I think the error here is ok because we are working in a concrete context in which there is no `first` ans `second` definitions. Current container knows nothing about them.

xepozz: But must

viktorprogger: For me it's like in OOP: your object can't access private fields of another object. If you want to get `first` inside a container, you must have `first` inside this container, not everywhere else. DI containers are made to decrease code coupling, and they musn't be tight coupled.

viktorprogger: The real problem I see is `CircularReferenceException` instead of `NotFoundException`. In this case container should throw `NotFoundException`.

xepozz: Ok, let's see the following case:
1. We have 2 containers in modules `moduleA` and `moduleB`, 1 composite container and 1 regular container from main app.
  Ultimately we have 1 aggregate container, that managed with 3 regular containers.
2. Regular container have definitions for all PSR interfaces: EventDispatcher, Cache, Http and etc.
3. Containers from `moduleA` and `moduleB` depends on PSR interfaces and don't configure any implementation.

How to pass to `moduleA` and `moduleB` our own implementations?

viktorprogger: `$moduleAContainer->addDefinition()`?
Maybe it's a good idea to add a definition class that can have a reference to another container.
Generally my point is that regular container must not resolve dependencies while they are not defined inside this particular container.

xepozz: `$moduleAContainer->addDefinition()` not PSR

samdark:
It is now fixed.

Issue #104, opened 2020-02-13
Incorrect work of FactoryProvider->has()


roxblnfk:
README.MD: # Using deferred service providers [[link](https://github.com/yiisoft/di#using-deferred-service-providers)]

> ```php
> // returns false as provider wasn't registered
> $container->has(EngineInterface::class); 
> 
> // returns SolarEngine, registered in the provider
> $engine = $container->get(EngineInterface::class); 
> 
> // returns true as provider was registered when EngineInterface was requested from the container
> $container->has(EngineInterface::class); 
> ```

PSR about `has()`:
> [1.1.2 Reading from a container](https://www.php-fig.org/psr/psr-11/#112-reading-from-a-container)
> has takes one unique parameter: an entry identifier, which MUST be a string. has MUST return true if an entry identifier is known to the container and false if it is not. If has($id) returns false, get($id) MUST throw a NotFoundExceptionInterface.

> [3.1. Psr\Container\ContainerInterface](https://www.php-fig.org/psr/psr-11/#31-psrcontainercontainerinterface)
> ```php
>     /**
>      * Returns true if the container can return an entry for the given identifier.
>      * Returns false otherwise.
>      *
>      * `has($id)` returning true does not mean that `get($id)` will not throw an exception.
>      * It does however mean that `get($id)` will not throw a `NotFoundExceptionInterface`.
>      *
>      * @param string $id Identifier of the entry to look for.
>      *
>      * @return bool
>      */
>     public function has($id);
> ```

Method `get()` must returns `true` if the container can return an entry for the given identifier. 
In our case, it returns `false` until the first call to `get()`

samdark:
Need to cover it with unit test...

PR #113, opened 2020-02-20
Improve di container

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #104

### DIFF
--- a/phpbench.json
+++ b/phpbench.json
@@ -1,6 +1,6 @@
 {
     "bootstrap": "vendor/autoload.php",
-    "path": "/tests/benchmark",
+    "path": "tests/benchmark",
     "retry_threshold": 5,
     "php_disable_ini": true,
     "php_config": {
--- a/phpbench.json
+++ b/phpbench.json
@@ -1,6 +1,6 @@
 {
     "bootstrap": "vendor/autoload.php",
-    "path": "tests/benchmark",
+    "path": "/tests/benchmark",
     "retry_threshold": 5,
     "php_disable_ini": true,
     "php_config": {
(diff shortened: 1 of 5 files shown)


phpbench.json
```diff
@@ -1,6 +1,6 @@
 {
     "bootstrap": "vendor/autoload.php",
-    "path": "tests/benchmark",
+    "path": "/tests/benchmark",
```
samdark: That's not correct. Should be reverted.

src/Contracts/DeferredServiceProviderInterface.php
```diff
@@ -16,4 +16,6 @@ interface DeferredServiceProviderInterface extends ServiceProviderInterface
      * @return bool whether service provider would register definition or not.
      */
     public function hasDefinitionFor(string $id): bool;
+
+    public function provides(): array;
```
samdark:
```suggestion
    /**
      * @return string[] a list of IDs of services provided
      */
    public function provides(): array;
```

samdark:
Thanks!

phpbench.json
yiiliveext: Pushed accidentally, otherwise it does not work on my PC.

samdark: That is weird.

yiiliveext: > That is weird.

No, just a phpbench bug, I will make PR for that.

yiiliveext: Ups... The bug is already fixed, I've just run composer update :)

PR #114, opened 2020-02-23
Fix benchmarks, add docs

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/tests/Benchmark/ContainerBench.php
+++ b/tests/Benchmark/ContainerBench.php
@@ -33,7 +33,7 @@ class ContainerBench
     private $indexes = [];
 
     /** @var int[] */
-    private $rundomIndexes = [];
+    private $randomIndexes = [];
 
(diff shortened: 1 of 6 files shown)


tests/Benchmark/ContainerBench.php
```diff
@@ -44,12 +58,14 @@ public function before(): void
     {
         $definitions = [];
         $definitions2 = [];
-        for ($i = 0; $i < 1000; $i++) {
+        for ($i = 0; $i < 200; $i++) {
```
xepozz: Can you use `200` as a constant to avoid errors "out of range"?

yiiliveext: Done

samdark:
👍

Issue #110, opened 2020-02-19
Feature: decorators


xepozz:
Would be nice allow to decorate already existing services in container.
I see at least 2 variants of usage:
1. Decorating already existing service that exist in another container (applied only to `CompositeContainer`)
  // config/common.php
  ```php
  return [
    // ...
    LoggerInterface::class => function() use ($compositeContainer) {
      return new LoggerDecorator($compositeContainer->get(LoggerInterface::class));
    }
    // ...
  ];
  ```
  At now it parsed as `circle dependency`, but we can modernize this syntax. At now it doesn't matter.
2. Decorating already existing services in current container
  // config/common.php
  ```php
  return [
    // ...
    LoggerInterface::class => /* ... */,
    
    LoggerDecorator::class => [
        'decorated' => LoggerInterface::class
    ];
    // ...
  ];
  ```
  or
// config/common.php
  ```php
  return [
    // ...
    LoggerInterface::class => /* ... */,
    // ...
  ];
  ```
  // config/common.dev.php
  ```php
  return [
    // ...
    LoggerInterface::class => function ($container) {
        return new LoggerDecorator($container->get(LoggerInterface::class));
    }
    // ...
  ];
  ```
   
### What is the expected result?
In all cases `$container->get(LoggerInterface::class)` should return `LoggerDecorator`

### What do you get instead?
In the first case we have `CircularReferenceException`. In the second case we have `LoggerInterface` implementation that passed via `/* ... */ or `CircularReferenceException`.

### Additional info
I think we need to think all future features for another cases (I'm not about decorators only).

xepozz:
I made syntax examples. 

```php
LoggerInterface::class => (new ContainerBuilder())
  ->callback(fn () => new Logger())
  ->decorate(fn ($container, $logger) => new LoggerDecorator($logger));
```

```php
LoggerInterface::class => (new ContainerBuilder())
  ->callback(fn () => new Logger())
  ->decorate(LoggerDecorator::class),
LoggerDecorator::class => fn ($container) => new LoggerDecorator($container->get(LoggerInterface::class)),
```

```php
$definitions = [
  LoggerInterface::class => fn () => new Logger(),
  LoggerDecorator::class => fn ($container) => new LoggerDecorator($container->get(LoggerInterface::class)),
];
$containerBuilder
  ->decorate(LoggerInterface::class, LoggerDecorator::class)
  ->decorate(ContainerInterface::class, ContainerDecorator::class);
```

[...]

PR #116, opened 2020-03-01
Add container proxy

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #107 #110

Don't merge! This is a demo PR. 

Old container syntax
```php
$container = new Container($definitions);
```
New container syntax 
```php
$container = Container::getBuilder()->registerDefinitions($definitions)->build();
```
index.php container&debugger registration expample
```php
$containerBuilder = Container::getBuilder()
    ->registerDefinitions(require Builder::path('web'));

/**
 * @var array $params The variable is available after requiring config files.
 */
$debugEnabled = (bool)($params['debugger.enabled'] ?? false) && class_exists(Debugger::class);
if ($debugEnabled) {
   $debuggerServiceProvider = new DebugServiceProvider();
   $containerBuilder->registerServiceProviders([$debuggerServiceProvider]);
}
$container = $containerBuilder->build();
```

Container proxy config
```php
// /config/web.php
CommonServiceCollectorInterface::class => CommonServiceCollector::class,
ContainerProxyInterface::class => static function (ContainerInterface $container) use ($params) {
        $collector = $container->get(CommonServiceCollectorInterface::class);
        $dispatcher = $container->get(EventDispatcherInterface::class);
        $debuggerEnabled = (bool)($params['debugger.enabled'] ?? false);
        $trackedServices = (array)($params['debugger.trackedServices'] ?? []);
        $decoratedServices = (array)($params['container.decorators'] ?? []);
        $runtime = $container->get(Aliases::class)->get('@runtime');
        $path = "$runtime/cache/container-proxy";
        if (!is_dir($path)) {
            mkdir($path);
        }
        $logLevel = ContainerProxy::LOG_ARGUMENTS | ContainerProxy::LOG_RESULT | ContainerProxy::LOG_ERROR;
        return new ContainerProxy(
            $debuggerEnabled,
            array_merge($trackedServices, $decoratedServices),
            $container,
            $dispatcher,
            $collector,
            $path,
            $logLevel
        );
},
```
```php
// /config/params.php
'debugger.enabled' => true,
'debugger.trackedServices' => [
    \Cycle\ORM\FactoryInterface::class,
    \Yiisoft\Mailer\MailerInterface::class,
    \Psr\Http\Message\ServerRequestFactoryInterface::class,
    \Yiisoft\Yii\Web\Session\SessionInterface::class,
    \Yiisoft\Router\RouteCollectorInterface::class,
],
'debugger.collectors' => [
    \Yiisoft\Di\CommonServiceCollector::class
],
'container.decorators' => [
    // custom method decorator
    \Yiisoft\Router\UrlGeneratorInterface::class => [
        'generate' =>function ($result, $name, $params = []) {
            if ($name === 'site/contact') {
                return $result .  '/me';
            }
            return $result;
        }
    ],
],
```

Also, you can remove debugger enable from the  index.php, and enable it from config
```php
'debugger.trackedServices' => [
    //array-configured interface decorator/proxy, the first param is a proxy/decorator class, other params are decorator's constructor parameters
    LoggerInterface::class => [LoggerInterfaceProxy::class, LogCollectorInterface::class],
    //callback-cofigured interface decorator/proxy
    EventDispatcherInterface::class => function (ContainerInterface $container) {
        $dispatcher = $container->get(EventDispatcherInterface::class);
        $compositeDispatcher = new CompositeDispatcher();
        $compositeDispatcher->attach($container->get(DebugStartupDispatcher::class));
        $compositeDispatcher->attach($container->get(DebugEventDispatcher::class));
        $compositeDispatcher->attach($dispatcher);
        $compositeDispatcher->attach($container->get(DebugShutdownDispatcher::class));

        return new EventDispatcherInterfaceProxy($compositeDispatcher, $container->get(EventCollectorInterface::class));
    },
],
```


composer.json
```diff
@@ -20,11 +20,15 @@
         "irc": "irc://irc.freenode.net/yii",
         "source": "https://github.com/yiisoft/di"
     },
+    "minimum-stability": "dev",
+    "prefer-stable": true,
     "require": {
         "php": "^7.4",
         "ext-mbstring": "*",
         "psr/container": "1.0.0",
+        "psr/event-dispatcher": "1.0.0",
```
samdark: That's a no go. DI container should not be aware of event dispatcher.

composer.json
```diff
@@ -20,11 +20,15 @@
         "irc": "irc://irc.freenode.net/yii",
         "source": "https://github.com/yiisoft/di"
     },
+    "minimum-stability": "dev",
```
xepozz: ?

src/CommonMethodProxy.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Yiisoft\Di;
+
+use Psr\EventDispatcher\EventDispatcherInterface;
+
+class CommonMethodProxy extends CommonServiceProxy
```
samdark: Was it planned to move it to debugger?

composer.json
yiiliveext: It's a temporary dependency

src/CommonMethodProxy.php
yiiliveext: I planned to move it to a separate package like `yiisoft/proxy` together with poxy classes from `yiisoft/yii-debug`.

composer.json
yiiliveext: These temporary changes will be removed.

yiiliveext: Done

composer.json
schmunk42: Why is it temporarily needed?

yiiliveext: > Why is it temporarily needed?

Because this PR contains classes that will be moved to other separate packages.

src/CommonMethodProxy.php
samdark: Why would we need concrete proxy classes if we'll have configurable universal proxy?

src/CompositeContainer.php
```diff
@@ -44,8 +44,14 @@ public function has($id)
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(ContainerInterface $container, bool $parentLookup = false): void
```
samdark: A common name for it is "delegate lookup".

```suggestion
    public function attach(ContainerInterface $container, bool $delegateLookup = false): void
```

Also, a phpdoc is needed. It should explain that when set to true, dependencies will be looked for in composite container instead of in $container directly. When set to false, dependencies would be looked for in $container only.

src/CompositeContainer.php
```diff
@@ -44,8 +44,14 @@ public function has($id)
...
     {
+        if ($parentLookup && $container instanceof Container) {
+            $container = (new ContainerBuilder($container->withParentContainer($this)))->build();
+        } elseif ($this->has(ContainerProxyInterface::class)) {
```
samdark: Should composite container take care about it? Isn't it better to do it like the following?

```php
$moduleContainer = new ModuleContainer($config);
$proxy = new ProxyContainer($moduleContainer);

$compositeContainer->attach($proxy);
```

src/Container.php
```diff
@@ -34,6 +34,8 @@ class Container implements ContainerInterface
      */
     private $instances;
 
+    private ?ContainerInterface $parentContainer = null;
```
samdark: In order to delegate lookup to work properly it should be always topmost container, not just parent.

src/CompositeContainer.php
samdark: Currently it wraps with proxy if proxy found in *any* sub-container. That doesn't sound like a stable thing. Any container may override it.

src/CompositeContainer.php
```diff
@@ -44,8 +45,18 @@ public function has($id)
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(ContainerInterface $container, bool $parentLookup = false, bool $enableProxy = true): void
     {
+        if ($parentLookup && $container instanceof Container) {
+            $container = (new ContainerBuilder($container->withParentContainer($this)))->build();
+        } elseif ($enableProxy && $this->has(ContainerProxyInterface::class)) {
```
samdark: Composite container should not be aware of proxy.

src/CompositeContainer.php
```diff
@@ -44,8 +45,18 @@ public function has($id)
      * Attaches a container to the composite container.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(ContainerInterface $container, bool $parentLookup = false, bool $enableProxy = true): void
```
samdark: See my comment about "delegate lookup". It's better to use that name...

src/CompositeContainer.php
samdark:
```php
$compositeContainer->attach(ProxyContainer::for($container));
```

src/Container.php
samdark: That's usually called `$rootContainer`.

src/CompositeContainer.php
samdark: See https://github.com/yiisoft/proxy/pull/2. That's my try on generic proxy. It doesn't need to know anything about container and doesn't require delegate lookup.

src/CompositeContainer.php
yiiliveext: Done

src/Container.php
yiiliveext: See last commit bead8aa

yiiliveext:
How to create proxy:

The first way (if ContainerProxyInterface is assigned in the container config) :
```php
$container = Container::getBuilder()->registerDefinitions($definitions)->build();
```
The second way (if ContainerProxyInterface is not assigned in the container config) :
```php
$container = new Container($definitions);
$container = new ContainerProxy($container, $globalContainer->get(ContainerProxyConfig::class));
```

src/CompositeContainer.php
yiiliveext: Done

src/CompositeContainer.php
```diff
@@ -42,10 +42,15 @@ public function has($id)
...
      * Attaches a container to the composite container.
+     * When $delegateLookup set to true, dependencies will be looked for in composite container instead of
+     * in $container directly. When set to false, dependencies would be looked for in $container only.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(ContainerInterface $container, bool $delegateLookup = false): void
     {
+        if ($delegateLookup && $container instanceof ContainerDelegateInterface) {
+            $container = $container->withRootContainer($this);
```
samdark: We can avoid composite container knowing about implementation detail of containers it wraps. It is more than likely that third part containers won't ever implement ContainerDelegateInterface...

```php
$composite = new CompositeContainer();

$container1 = new Container($config1, $composite);
$composite->attach($container1);

$container2 = AnotherContainer::fromArray($config2)->delegateLookupTo($comosite);
$composite->attach($container2);
```

src/Container.php
```diff
@@ -48,9 +50,22 @@ public function __construct(
         array $providers = []
     ) {
         $this->setMultiple($definitions);
-        foreach ($providers as $provider) {
-            $this->addProvider($provider);
+        $this->registerServiceProviders($providers);
+    }
+
+    public function withRootContainer(ContainerInterface $container): ContainerInterface
```
samdark: Root container is better to be passed to constructor.

src/Container.php
```diff
@@ -48,9 +50,22 @@ public function __construct(
...
+    {
+        $newContainer = clone $this;
+
+        if ($this->rootContainer === null) {
+            $newContainer->rootContainer = new CompositeContainer();
```
samdark: Is there a need to create root container like that?

src/Container.php
```diff
@@ -179,6 +193,13 @@ public function setMultiple(array $config): void
         }
     }
 
+    public function registerServiceProviders(array $providers): void
```
samdark:
```suggestion
    public function addProviders(array $providers): void
```

src/ContainerBuilder.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+namespace Yiisoft\Di;
+
+use Psr\Container\ContainerInterface;
+use Psr\Container\ContainerExceptionInterface;
+use Yiisoft\Factory\Factory;
+
+final class ContainerBuilder
```
samdark: I don't think it is a good idea to have it in this package.

1. It doesn't add much convenience.
2. It bounds container to `ContainerProxyInterface`.

src/ContainerProxyInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Yiisoft\Di;
+
+use Psr\Container\ContainerInterface;
+
+interface ContainerProxyInterface extends ContainerInterface
```
samdark: Is it debug specific? If so, let's move it there. Later it could be moved somewhere if there will be demand for it.

samdark:
https://github.com/yiisoft/di/pull/109 could be merged into it to validate issue is fixed.

src/Container.php
yiiliveext: There is no another way to do that.

src/Container.php
yiiliveext: It is not necessary at the moment of the container creation.

src/CompositeContainer.php
yiiliveext: It is for the native container only.

src/ContainerBuilder.php
yiiliveext: I prepared ContainerBuilder to another package.

src/ContainerProxyInterface.php
yiiliveext: see above

src/Container.php

src/Container.php
samdark: Can't you leave it as null?

src/ContainerBuilder.php
samdark: Let's do it then.

yiiliveext: Create a separate package like `yiisoft/container-bulder`

src/Container.php
yiiliveext: No, I can't. If you have called this method you want to set a root container.

src/CompositeContainer.php
samdark: But the whole idea of composite container is that it can work with any container implementation.

src/ContainerBuilder.php
samdark: `yiisoft/container-proxy`?

samdark: Would it be used anywhere outside of debug?

src/CompositeContainer.php
yiiliveext: Nothing has changed. We still can use any PSR-11 containers.

src/ContainerBuilder.php
yiiliveext: Yes, I'll use it for automated debugger&proxy enable|disable.

src/CompositeContainer.php
```diff
@@ -42,9 +42,11 @@ public function has($id)
 
     /**
      * Attaches a container to the composite container.
+     * When $delegateLookup set to true, dependencies will be looked for in composite container instead of
+     * in $container directly. When set to false, dependencies would be looked for in $container only.
      * @param ContainerInterface $container
      */
-    public function attach(ContainerInterface $container): void
+    public function attach(ContainerInterface $container, bool $delegateLookup = false): void
```
samdark: `$delegateLookup` could be removed now.

src/Container.php
```diff
@@ -48,9 +50,22 @@ public function __construct(
         array $providers = []
     ) {
         $this->setMultiple($definitions);
-        foreach ($providers as $provider) {
-            $this->addProvider($provider);
+        $this->addProviders($providers);
+    }
+
+    public function delegateLookup(ContainerInterface $container): ContainerInterface
```
samdark: If there's a way not to have it, I'd prefer not adding it.

src/AbstractContainerConfigurator.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Yiisoft\Di;
+
+abstract class AbstractContainerConfigurator
+{
+    protected function set(string $id, $definition): void
```
xepozz: Maybe `abstract protected function set(string $id, $definition): void;` instead?

samdark: No, that would cause errors.

PR #119, opened 2020-03-11
Add conditional singletons

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Conditional singletons. When the `__cacheCondition` equal true, the cached singleton instance will be returned, otherwise, the cache will be reset and a new instance will be returned 

Examples:
This is a session-dependent service. When a new session starts, a new instance of the `IdentityRepository` will be returned.  
```php
IdentityRepositoryInterface::class => [
        '__definition' => static function (ContainerInterface $container) {
            return $container->get(\Cycle\ORM\ORMInterface::class)->getRepository(\App\Entity\User::class);
        },
        '__cacheCondition' => function (ContainerInterface $container) {
                $currentSession = $container->get(SessionInterface::class);
                $lastSessionStorage = $container->get(\App\LastSessionStorage::class);
                $sameSession = ($currentSession->getId() === $lastSessionStorage->getSession()->getId());
                if (!$sameSession) {
                    $lastSessionStorage->setSession($currentSession);
                }

                return $sameSession;
            },
    ],
```
This is not a singleton service. Every time a new instance will be created.
```php
MyNotSingletonServiceInterface::class => [
    '__class' => IAmNotSingleton::class,
   '__cacheCondition' => false,
   '__construct()' => ['someParam']
]
```

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -51,7 +51,8 @@ final class Container extends AbstractContainerConfigurator implements Container
         array $definitions = [],
         array $providers = [],
         ContainerInterface $rootContainer = null
-    ) {
+    )
+    {
         $this->setMultiple($definitions);
         $this->addProviders($providers);
         if ($rootContainer !== null) {
@@ -171,14 +172,18 @@ final class Container extends AbstractContainerConfigurator implements Container
         }
     }
 
-    private function getCacheCondition(array $definition): array
+    private function getCacheCondition(array $definition)
     {
         return $definition['__cacheCondition'];
     }
 
     private function hasCacheCondition($definition): bool
     {
-        if (is_array($definition) && isset($definition['__cacheCondition']) && is_array($definition['__cacheCondition'])) {
+        if (is_array($definition) && isset($definition['__cacheCondition'])) {
+            if (!(is_bool($definition['__cacheCondition']) || is_callable($definition['__cacheCondition']))) {
+                throw new \RuntimeException("Definition array key '__cacheCondition' must be boolean or callable.");
+            }
+
             return true;
         }
 
@@ -198,19 +203,12 @@ final class Container extends AbstractContainerConfigurator implements Container
     private function validateCache($id): void
     {
         if (isset($this->cacheConditions[$id])) {
-            foreach ($this->cacheConditions[$id] as $service => $dependency) {
-                $currentValue = $service;
-                $dependencyValue = $dependency;
-                if (is_callable($dependency)) {
-                    $dependencyValue = $dependency($this);
-                }
-                if ($this->has($service)) {
-                    $currentValue = $this->get($service);
-                }
-                if ($currentValue != $dependencyValue) {
-                    unset($this->instances[$id]);
-                    break;
-                }
+            $condition = $this->cacheConditions[$id];
+            if (is_callable($condition)) {
+                $condition = $condition($this);
+            }
+            if (!$condition) {
+                unset($this->instances[$id]);
             }
         }
     }
--- a/src/Container.php
+++ b/src/Container.php
@@ -51,8 +51,7 @@ final class Container extends AbstractContainerConfigurator implements Container
         array $definitions = [],
         array $providers = [],
         ContainerInterface $rootContainer = null
-    )
-    {
+    ) {
         $this->setMultiple($definitions);
         $this->addProviders($providers);
         if ($rootContainer !== null) {
@@ -87,7 +86,7 @@ final class Container extends AbstractContainerConfigurator implements Container
     public function get($id, array $parameters = [])
     {
         $id = $this->getId($id);
-        $this->validateCache($id);
+        $this->invalidateCache($id);
         if (!isset($this->instances[$id])) {
             $this->instances[$id] = $this->build($id, $parameters);
         }
@@ -200,7 +199,7 @@ final class Container extends AbstractContainerConfigurator implements Container
         return $definition;
     }
 
-    private function validateCache($id): void
+    private function invalidateCache($id): void
     {
         if (isset($this->cacheConditions[$id])) {
             $condition = $this->cacheConditions[$id];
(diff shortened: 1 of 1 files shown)


src/Container.php
```diff
@@ -164,6 +171,50 @@ protected function processDefinition($definition): void
...
+            foreach ($this->cacheConditions[$id] as $service => $dependency) {
+                $currentValue = $service;
+                $dependencyValue = $dependency;
+                if (is_callable($dependency)) {
+                    $dependencyValue = $dependency($this);
+                }
+                if ($this->has($service)) {
+                    $currentValue = $this->get($service);
+                }
+                if ($currentValue != $dependencyValue) {
```
viktorprogger:
```suggestion
                if ($currentValue !== $dependencyValue) {
```
It's simple to use equal values on the left and on the right side. On the other hand there are lots of errors based on type conversion.

yiiliveext: It makes sense. But this is just a draft for now.

yiiliveext: An array key can't be boolean, it is converted to an integer value. So, if we want to use `strong equal` and the boolean key, the right part must returns `1` or `0` only.

src/Container.php
```diff
@@ -164,6 +171,50 @@ protected function processDefinition($definition): void
...
     }
 
+    private function getCacheCondition(array $definition): array
+    {
+        return $definition['__cacheCondition'];
+    }
+
+    private function hasCacheCondition($definition): bool
+    {
+        if (is_array($definition) && isset($definition['__cacheCondition']) && is_array($definition['__cacheCondition'])) {
```
xepozz: Looks like simple condition, that can be write by one-line: `return is_array() && isset()...`
And we can hack it with the short form `return is_array($definition['__cacheCondition'] ?? false)` :)

src/Container.php
```diff
@@ -164,6 +171,50 @@ protected function processDefinition($definition): void
...
+            return true;
+        }
+
+        return false;
+    }
+
+    private function clearDefinitionCacheCondition(array $definition)
+    {
+        unset($definition['__cacheCondition']);
+        if (isset($definition['__definition'])) {
```
xepozz: Is it used anywhere?

src/Container.php
yiiliveext: Maybe it would be better to use smth like
```php
'__cacheCondition' => function (ContainerInterface $container) {
    //complex condition calculation
    return (bool)$boolResult;
}
```

src/Container.php
yiiliveext: Yeah, when this method is called.

src/Container.php
xepozz: Yes, right

src/Container.php
xepozz: So is it redundant?

yiiliveext: I misunderstood you. It is not used in the array definition.

src/Container.php

src/Container.php

src/Container.php
```diff
@@ -164,6 +171,47 @@ protected function processDefinition($definition): void
...
 
+    private function getCacheCondition(array $definition)
+    {
+        return $definition['__cacheCondition'];
+    }
+
+    private function hasCacheCondition($definition): bool
+    {
+        if (is_array($definition) && isset($definition['__cacheCondition'])) {
+            if (!(is_bool($definition['__cacheCondition']) || is_callable($definition['__cacheCondition']))) {
```
xepozz: > is_bool($definition['__cacheCondition'])

In this case this is not a service so it shouldn't be declared in the container definitions.

xepozz:
How about to write a few test cases? :)

src/Container.php
yiiliveext: See the first post examples.

samdark:
That's interesting.

Are you sure multiple conditions and conditions overall are needed? For [RoadRunner](https://github.com/yiisoft/docs/blob/master/guide/en/tutorial/using-yii-with-roadrunner.md) / [Swoole](https://github.com/yiisoft/docs/blob/master/guide/en/tutorial/using-yii-with-swoole.md) we have entry points that clearly define the workflow such as:

```php
// RR and app init
$statelessContainer = new Container(require Builder::path('stateless'));
$statefulContainer = new Container(require Builder::path('stateful'));

$container = new CompositeContainer();
$container->attach($statelessContainer);
$container->attach($statefulContainer);

while ($request = $psr7->acceptRequest()) {
    try {
        $response = $application->handle($request);
        $psr7->respond($response);
    } catch (\Throwable $e) {
        $psr7->getWorker()->error((string)$e);
    }
    gc_collect_cycles();
    $statefulContainer->reset(); // <--- THIS
}

$application->shutdown();
```

yiiliveext:
> That's interesting.
> 
> Are you sure multiple conditions and conditions overall are needed? For [RoadRunner](https://github.com/yiisoft/docs/blob/master/guide/en/tutorial/using-yii-with-roadrunner.md) / [Swoole](https://github.com/yiisoft/docs/blob/master/guide/en/tutorial/using-yii-with-swoole.md) we have entry points that may clearly define the workflow such as:

Your solution is for the main container only. Every module/package/ service can have its own container. In this way (this PR) you can set session-dependent or request-dependent singletons in any container.

src/Container.php
xepozz: Are you about the second example?
This is not a service

yiiliveext: Why?

xepozz: Because in this case `Container` will work as a factory.
Thus to use the factory pattern to make these classes will be better solution imho.

Also your second case isn't real

samdark:
Closed in favor of https://github.com/yiisoft/di/pull/120

PR #127, opened 2020-06-16
Cleanup

samdark:
- Add types
- Add strict types
- Fix leftovers in benchmarks
- Remove unused imports
- Remove unused phpdoc
- Correct phpdoc mistakes
- Correct spelling mistakes

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Yiisoft\Di;
 
 use Psr\Container\ContainerInterface;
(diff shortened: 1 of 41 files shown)


src/Container.php
```diff
@@ -22,28 +24,29 @@ final class Container extends AbstractContainerConfigurator implements Container
...
 
     private ?ContainerInterface $rootContainer = null;
 
     /**
      * Container constructor.
      *
-     * @param array $definitions
-     * @param ServiceProviderInterface[] $providers
+     * @param array $definitions Definitions to put into container.
+     * @param ServiceProviderInterface[]|array $providers Service providers to get definitions from.
```
roxblnfk, line 45:
```suggestion
     * @param ServiceProviderInterface[] $providers Service providers to get definitions from.
```

src/Container.php
```diff
@@ -22,28 +24,29 @@ final class Container extends AbstractContainerConfigurator implements Container
...
      * to detect circular references
      */
-    private $building = [];
+    private array $building = [];
 
     /**
      * @var object[]
      */
-    private $instances;
+    private array $instances;
```
roxblnfk, line 37:
```suggestion
    private array $instances = [];
```

src/Container.php
samdark: There is a concept of deferred providers and these are specified as class names. See https://github.com/yiisoft/di/blob/master/tests/Unit/DeferredServiceProviderTest.php#L28

roxblnfk: hm... `ServiceProviderInterface[]|string[]` ?

samdark: Yes. I'll change it. Thanks.

PR #134, opened 2020-06-28
Fix composite container

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -19,13 +19,15 @@ final class CompositeContainer implements ContainerInterface
      */
     private array $containers = [];
 
-    public function get($id, array $parameters = [])
+    public function get($id)
     {
         foreach ($this->containers as $container) {
+            if ($container->has($id, true)) {
+                return $container->get($id);
+            }
+        }
+        foreach ($this->containers as $container) {
             if ($container->has($id)) {
-                if ($parameters !== []) {
-                    return $container->get($id, $parameters);
-                }
                 return $container->get($id);
             }
         }
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -11,7 +11,7 @@ use Yiisoft\Factory\Exceptions\NotFoundException;
  * This class implements a composite container for use with containers that support the delegate lookup feature.
  * The goal of the implementation is simplicity.
  */
-final class CompositeContainer implements ContainerInterface
+final class CompositeContainer extends AbstractContainerConfigurator implements ContainerInterface
 {
     /**
      * Containers to look into starting from the beginning of the array.
@@ -21,6 +21,12 @@ final class CompositeContainer implements ContainerInterface
 
     public function get($id)
     {
+        foreach ($this->containers as $container) {
+            if (method_exists($container, 'hasDefinition') && $container->hasDefinition($id)) {
+                return $container->get($id);
+            }
+        }
+
         foreach ($this->containers as $container) {
             if ($container->has($id)) {
                 return $container->get($id);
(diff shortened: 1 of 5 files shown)


src/CompositeContainer.php
```diff
@@ -19,13 +19,15 @@ final class CompositeContainer implements ContainerInterface
      */
     private array $containers = [];
 
-    public function get($id, array $parameters = [])
+    public function get($id)
     {
+        foreach ($this->containers as $container) {
+            if ($container->has($id, true)) {
```
viktorprogger, line 25: What if here will be non-yiisoft container?

viktorprogger:
```suggestion
            if (method_exists($container, 'hasDefinition') && $container->hasDefinition($id)) {
```

yiiliveext: > What will happen if here will be non-yiisoft container?

See over league container test, the second param will be ignored

viktorprogger: Anyway extra parameter usage is a bad practice when there is another way to do what you want to do. The Yiisoft and League containers are not the only implementations of this PSR, and there is no necessity to make our implementation incompatible with some others.
I suggest to add one more interface (e.g. `DefinitionVerifiableInterface`) and check if the container implements it (instead of `method_exists`). This way wel'll kill two birds with one stone: 
- functionality will be achieved 
- compatibility with will be saved

yiiliveext: > Anyway extra parameter usage is a bad practice when there is another way to do what you want to do. The Yiisoft and League containers are not the only implementations of this PSR, and there is no necessity to make our implementation incompatible with some others.
> I suggest to add one more interface (e.g. `DefinitionVerifiableInterface`) and check if the container implements it (instead of `method_exists`). This way we'll kill two birds with one stone:
> 
> * functionality will be achieved
> * compatibility with the PSR will be saved

It leads to the container interface violation

viktorprogger: > It leads to the container interface violation

Your way leads to the interface violation while adding one more interface is expansion of area of responsibility of the concrete implementation (so many "of" :sweat_smile:).

samdark:
It fixes the following case:

```
$compositeContainer
  $subContainer 1
  $subContainer 2
    MyClass => [ ... ]

$myClass = $compositeContainer->get(MyClass::class);
```

Currently we'll get new instance of MyClass without config applied because Yii DI container tries to instantiate a class in case it's not in the config.

src/Container.php
```diff
@@ -66,9 +66,9 @@ public function __construct(
      * @return bool whether the container is able to provide instance of class specified.
      * @see set()
      */
-    public function has($id): bool
+    public function has($id, bool $strict = false): bool
```
samdark, line 69: I suggest moving `$strict` to container constructor / setter. Then there will be no need to expand interface signature.

tests/Unit/AbstractCompositePsrContainerTest.php
```diff
@@ -70,5 +70,14 @@ public function testGetPriority(): void
         $compositeContainer->attach($containerOne);
         $compositeContainer->attach($containerTwo);
         $this->assertInstanceOf(EngineMarkTwo::class, $compositeContainer->get(EngineInterface::class));
+
+        $containerOne = new Container([EngineMarkOne::class => [
+            '__class' => EngineMarkTwo::class,
+        ]]);
+        $containerTwo = new Container([], [], null, true);
```
viktorprogger:
```suggestion
        $containerOne = new Container();
        $containerTwo = new Container([EngineMarkOne::class => EngineMarkTwo::class], [], null, true);
```

viktorprogger:
Moved the solution with strict mode to #135

samdark:
The use-case is modules. When we're within module, we're using its container that, if there's no dependency, delegates lookup to application container.

PR #135, opened 2020-06-30
Fix prioritizing dependencies

viktorprogger:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | ❌

Based on the early #134.

> It fixes the following case:
> 
> ```
> $compositeContainer
>   $subContainer 1
>   $subContainer 2
>     MyClass => [ ... ]
> 
> $myClass = $compositeContainer->get(MyClass::class);
> ```
> 
> Currently we'll get new instance of MyClass without config applied because Yii DI container tries to instantiate a class in case it's not in the config.

© @samdark

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -89,11 +89,12 @@ final class Container extends AbstractContainerConfigurator implements Container
      */
     public function get($id)
     {
+        $id = $this->getId($id);
+
         if (!$this->has($id)) {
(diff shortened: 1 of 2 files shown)


samdark:
Won't work conceptually as we agreed in the chat.

PR #139, opened 2020-08-04
Fix tests php 8.0.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

- [x] add workflok benchmark.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "irc": "irc://irc.freenode.net/yii",
         "source": "https://github.com/yiisoft/di"
     },
+    "minimum-stability": "dev",
     "require": {
         "php": "^7.4",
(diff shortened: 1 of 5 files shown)


samdark:
Running benchmark in CI is awesome idea!

composer.json
```diff
@@ -20,6 +20,7 @@
         "irc": "irc://irc.freenode.net/yii",
         "source": "https://github.com/yiisoft/di"
     },
+    "minimum-stability": "dev",
```
samdark, line 23: Is that required for phpunit?

terabytesoftw: yes.

samdark:
Thanks!

PR #143, opened 2020-09-01
Change providers strategy

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️


src/AbstractContainerConfigurator.php
```diff
@@ -8,16 +8,6 @@
 
 abstract class AbstractContainerConfigurator
```
samdark, line 9: Seems it's near to empty now or at least name doesn't reflect the purpose anyway.

src/Container.php
```diff
@@ -267,7 +277,8 @@ private function addProvider($providerDefinition): void
      */
     private function buildProvider($providerDefinition): ServiceProviderInterface
     {
-        $provider = Normalizer::normalize($providerDefinition)->resolve($this);
+
+        $provider = $this->factory->create($providerDefinition);
```
samdark, line 281: Why do we need to create an object right away? Can't we set the config as is?

samdark (changes requested):

src/Container.php
yiiliveext: > Why do we need to create an object right away? Can't we set the config as is?

We need to set params in the provider constructor.

yiiliveext: https://github.com/yiisoft/yii-event/blob/master/src/EventDispatcherProvider.php#L24

samdark: Discussed in private. Decided that @yiiliveext will adjust it so providers array is returning instances instead of definitions.

src/Container.php
```diff
@@ -237,43 +244,30 @@ private function addProviders(array $providers): void
...
+    private function addProvider($provider): void
     {
-        $provider = $this->buildProvider($providerDefinition);
+        if (!($provider instanceof ServiceProviderInterface)) {
+            throw new InvalidConfigException(
+                'Service provider should be an instance of ' . ServiceProviderInterface::class
+            );
+        }
 
         if ($provider instanceof DeferredServiceProviderInterface) {
```
BoShurik, line 262: Does `DeferredServiceProviderInterface` make sense now? Looks like all providers are lazy now

samdark: @yiiliveext looks like another leftover.

yiiliveext: DeferredProvider is for a heavyweight service initialization. For example, we should make a heavyweight API call and pass the received data to the service constructor. If we use this service only on a few routes, then why do we make the API call on each route? So, we use deferred providers to avoid this.

vjik:
**Suggestion.**

Add support `\Closure` for providers.
This is can use for not yii-* packages with simple code in providers.

For example, config:

```php
return [
    'yiisoft/csrf' => static function() {
    CsrfToken::initialize();
    return [];
  },
];
```

src/Container.php
```diff
@@ -237,43 +244,30 @@ private function addProviders(array $providers): void
...
-     * @param string|array $providerDefinition
+     * @param object $provider
      *
      * @throws InvalidConfigException
      * @throws NotInstantiableException
      * @see ServiceProviderInterface
      * @see DeferredServiceProviderInterface
      */
-    private function addProvider($providerDefinition): void
+    private function addProvider($provider): void
```
viktorprogger, line 254:
```suggestion
    private function addProvider(ServiceProviderInterface $provider): void
```

src/Container.php
BoShurik: @yiiliveext 
```php
return [
    HeavyService::class => static function (ContainerInterface $container) {
        $data = $container->get(Api::class)->heavyCall();
        
        return new HeavyService($data);
    },
];
```
Isn't this service will be initialized only when requested?

yiiliveext: > Isn't this service will be initialized only when requested?

Yes, this is another way, it is appropriate in a simple case, but we may have something like that
```php
class Provider implements ServiceProviderInterface
{
     public function getDefinitions(): array
     {
           $rawData = file_get_contents('http:\\mysyte.com\something');
           $dataForFirstService = $this->calculateDataForFirstService($rawData); 
           $dataForSecondService = $this->calculateDataForSecondService($rawData); 
           $dataForThirdService = $this->calculateDataForThirdService($rawData); 

           return [
               FirstService::class => [
                   '__construct()' => [$dataForFirstService],
               ],
               SecondService::class => [
                   '__construct()' => [$dataForSecondService],
               ],
               ThirdService::class => [
                   '__construct()' => [$dataForThirdService],
               ], 
           ]; 
     }

     public function calculateDataForFirstService($data)
     {
          //calculation
     }

     public function calculateDataForSecondService($data)
     {
          //calculation
     }

     public function calculateDataForThirdService($data)
     {
          //calculation
     }     
}
```

BoShurik: @yiiliveext make sense, but can be defined with factories:
```php
return [
    Api::class => [],
    ApiInterface::class => static function (ContainerInterface $container) {
        // To cache requests
        return new CachedApiDecorator($container->get(Api::class));
    },
    ServiceFactory::class => [
        // Also request may be cached in factory
        '__construct' => [Reference::to(ApiInterface::class)],
    ],
    FirstService::class => static function (ContainerInterface $container) {
        return $container->get(ServiceFactory::class)->createFirst();
    },
    SecondService::class => static function (ContainerInterface $container) {
        return $container->get(ServiceFactory::class)->createSecond();
    },
    ThirdService::class => static function (ContainerInterface $container) {
        return $container->get(ServiceFactory::class)->createThird();
    },
];
```

Anyway, in advanced cases you probably need additional data from container to access `API` (tokens, request signer etc)

yiiliveext: > Anyway, in advanced cases you probably need additional data from container to access `API` (tokens, request signer etc)

Yes, I think it would be fine to add `setContainer()` method to `DeferredProviderInterface`.

samdark:
1. Need to try that on yii-demo and packages. @yiiliveext will do it.
2. Need to decide if that's a good way to go overall.

samdark:
@yiiliveext seems https://github.com/yiisoft/widget/issues/12 will be broken after the change. How would solve it?

yiiliveext:
> @yiiliveext seems [yiisoft/widget#12](https://github.com/yiisoft/widget/issues/12) will be broken after the change. How would solve it?

Maybe it is BootstrapMiddleware in the main pipeline?

samdark:
Yeah, that could be solved that way.

samdark:
Here's a draft of bootstrapping: https://github.com/yiisoft/app/pull/76

samdark:
Won't be merged in this variant.

Issue #125, opened 2020-06-05
[RFC] Tags support


BoShurik:
E.g. I have a telegram bot with commands implementing some `CommandInterface`. And I have  `HelpCommand` which must show all available commands.
Now to achieve this I must edit `HelpCommand` configuration every time I add new command to the bot.
Using tags I can tag command as a `telegram-bot-command` so It will be automatically loaded by `HelpCommand`

Same feature in [Laravel](https://laravel.com/docs/7.x/container#tagging) and [Symfony](https://symfony.com/doc/current/service_container/tags.html#reference-tagged-services)

samdark:
i.e. Container collects all the tagged classes, forms a list and passed the list as an argument of `HelpCommand`.

samdark:
That could be useful but there are possible disadvantages as well: implicitness, unable to know what's passed to `HelpCommand` by looking the config.

BoShurik:
> unable to know what's passed to HelpCommand by looking the config

Catch me if I wrong, but in case non-FQCN service names it's almost the same level of knowledge:
```
return [
    SomeService::class => [
        '__construct()' => [Reference::to('some-dependency')],
    ],
]
```
and
```
return [
    HelpCommand::class => [
        '__construct()' => [Reference::tag('telegram-bot-command)],
    ],
]
```
In the first example you need to find `some-dependency` definition in all sources.

Of course it would be good to have some sort of debug tool. Something like Symfony's
`bin/console debug:container --tag='console.command'`

samdark:
Without tags it will be:

```php
return [
    SomeService::class => [
        '__construct()' => [
            Reference::to(Command1::class),
            Reference::to(Command2::class),
            Reference::to(Command3::class),
        ],
    ],
]
```

BoShurik:
I mean without FQCN services it would be
```
return [
    SomeService::class => [
        '__construct()' => [
            Reference::to('command'),
            Reference::to('other-command'),
            Reference::to('some-other-command'),
        ],
    ],
]
```
So you have to use grep to find class names. Almost the same as with tags

samdark:
We are usually using FQNs.

BoShurik:
Yes, everybody does, but sometimes it's impossible.
Anyway, it's hard to implement described example without tags.
The problem is that developers focusing on creating new commands and can forgot / don't know to add it in `HelpCommand`.

samdark:
[...]

PR #144, opened 2020-09-01
Add tags to container

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues | #125


src/Container.php
```diff
@@ -115,6 +119,8 @@ protected function delegateLookup(ContainerInterface $container): void
     protected function set(string $id, $definition): void
     {
         $this->validateDefinition($definition);
+        $tags = $this->extractTags($definition);
```
thenotsoft, line 122:
```suggestion
```

src/Container.php
```diff
@@ -115,6 +119,8 @@ protected function delegateLookup(ContainerInterface $container): void
...
+        $this->setTags($id, $tags);
```
thenotsoft, line 123:
```suggestion
        $this->setTags($id, $this->extractTags($definition));
```

yiiliveext: oudated

src/Container.php
yiiliveext: outdated

src/Container.php
```diff
@@ -131,18 +138,61 @@ protected function setMultiple(array $config): void
         }
     }
 
+    private function extractDefinition($definition)
+    {
+        if (is_array($definition) && isset($definition['__definition'])) {
```
samdark, line 143: Do we really need this new syntax?

```php
[
  '__definition' => [
      '__class' => MyClass::class,
      '__construct()' => [42],
  ],
  '__tags' => ['debug', 'users']
]
```

Isn't `__tags` without `__definition` enough?

src/Container.php
```diff
@@ -131,18 +138,61 @@ protected function setMultiple(array $config): void
...
+        }
+
+        return $definition;
+    }
+
+    private function extractTags($definition): array
+    {
+        if (is_array($definition) && isset($definition['__tags']) && is_array($definition['__tags'])) {
+            $tags = $definition['__tags'];
+            unset($definition['__tags']);
```
samdark, line 154: Why do we need to unset here?

src/Container.php
```diff
@@ -158,6 +208,25 @@ private function build(string $id)
         return $object;
     }
 
+    private function isTagAlias(string $id): bool
+    {
+        return substr($id, 0, 4) === 'tag@';
```
samdark, line 213:
```suggestion
        return strpos($id, 'tag@') === 0;
```

samdark (approved):
The idea itself is alright 👍 Need at least some docs to merge it.

src/Container.php
BoShurik: Maybe use `!tag` to be sure there is no such method

yiiliveext: > Do we really need this new syntax?
> 
> ```
> [
>   '__definition' => [
>       '__class' => MyClass::class,
>       '__construct()' => [42],
>   ],
>   '__tags' => ['debug', 'users']
> ]
> ```
> 
> Isn't `__tags` without `__definition` enough?

This is for closure definitions

samdark:
Need docs to get an idea about how it is intended to be used.

src/Container.php
yiiliveext: Yes, this is for array definitions
```php
[
   '__class' => MyClass::class,
   '__construct()' => [42],
   '__tags' => ['debug', 'users']
]
```

src/Container.php
xepozz: How about to make the second argument like a stub Definition and in closure we  can set all definitions properties like the following:
```php
Interface::class => function ($container, $definition) {
  $definition->setTags($tags);
  $definition->setAnythingElse($var);
  return new Class();
}
```

xepozz (commented):
What's the result will be for calling `$container->has('tag@engine')`?

yiiliveext:
> What's the result will be for calling `$container->has('tag@engine')`?

Good question.

src/Container.php
samdark: Interesting idea but needs to be considered separately. An issue would be helpful.

samdark: > This is for closure definitions

Huh?

src/Container.php
samdark: But the definition isn't used afterwards and there's no sense in removing anything from it.

yiiliveext: It's used afterward, why not?

samdark:
```php
private function extractTags($definition): array
{
    if (is_array($definition) && isset($definition['__tags']) && is_array($definition['__tags'])) {
        $tags = $definition['__tags'];
        unset($definition['__tags']);
	// definition isn't returned or used afterwards
        $this->checkTags($tags);
        return $tags;
    }
    return [];
}
```

yiiliveext: Oh, yes. Just in the previous version `$definition` was by reference)

xepozz:
Will a composite container collect all definitions from subcontainers?

src/Container.php
yiiliveext: Вщту

src/Container.php

src/Container.php
yiiliveext: > > This is for closure definitions
> 
> Huh?

```php
EngineMarkTwo::class => [
                '__definition' => function () {
                    return new EngineMarkTwo();
                },
                '__tags' => ['engine'],
            ]
```

yiiliveext:
> Will a composite container collect all tags definitions from subcontainers?

Done

samdark:
Ready for review. Docs are needed.

hiqsol:
Please see alternative proposal https://github.com/yiisoft/di/pull/152
It provides almostly the same but is much simpler.

samdark:
@yiiliveext would you please resolve conflicts?

yiiliveext:
> @yiiliveext would you please resolve conflicts?

Done

samdark (approved):

PR #145, opened 2020-09-02
Add module container test

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/tests/Unit/AbstractModuleContainerTest.php
+++ b/tests/Unit/AbstractModuleContainerTest.php
@@ -1,0 +1,153 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Di\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Yiisoft\Di\Tests\Support\Car;
+use Yiisoft\Di\Tests\Support\CarInterface;
+use Yiisoft\Di\Tests\Support\EngineInterface;
+use Yiisoft\Di\Tests\Support\EngineMarkOne;
+use Yiisoft\Di\Tests\Support\EngineMarkTwo;
+use Yiisoft\Factory\Definitions\Reference;
+
+abstract class AbstractModuleContainerTest extends TestCase
+{
+    abstract public function createModuleContainer(array $rootDefinitions, array $moduleDefinitions): array;
+
+    public function testClassExistsIgnore(): void
+    {
+        [, $moduleContainer] = $this->createModuleContainer([
+            Car::class => [
+                '__construct()' => [Reference::to(EngineMarkOne::class)]
+            ],
+        ], []);
+
+        $car = $moduleContainer->get(Car::class);
+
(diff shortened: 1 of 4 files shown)


tests/Unit/AbstractModuleContainerTest.php
```diff
@@ -0,0 +1,153 @@
...
+use Yiisoft\Di\Tests\Support\Car;
+use Yiisoft\Di\Tests\Support\CarInterface;
+use Yiisoft\Di\Tests\Support\EngineInterface;
+use Yiisoft\Di\Tests\Support\EngineMarkOne;
+use Yiisoft\Di\Tests\Support\EngineMarkTwo;
+use Yiisoft\Factory\Definitions\Reference;
+
+abstract class AbstractModuleContainerTest extends TestCase
+{
+    abstract public function createModuleContainer(array $rootDefinitions, array $moduleDefinitions): array;
```
samdark, line 17: Are there going to be multiple implementations?

tests/Unit/AbstractModuleContainerTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Yiisoft\Di\Tests\Support\Car;
+use Yiisoft\Di\Tests\Support\CarInterface;
+use Yiisoft\Di\Tests\Support\EngineInterface;
+use Yiisoft\Di\Tests\Support\EngineMarkOne;
+use Yiisoft\Di\Tests\Support\EngineMarkTwo;
+use Yiisoft\Factory\Definitions\Reference;
+
+abstract class AbstractModuleContainerTest extends TestCase
```
samdark, line 15: Tests in this class are correct but it looks like some are duplicating each other.

tests/Unit/ModuleContainerTest.php
```diff
@@ -0,0 +1,23 @@
...
+use Psr\Container\ContainerInterface;
+use Yiisoft\Di\Container;
+use Yiisoft\Di\CompositeContainer;
+
+class ModuleContainerTest extends AbstractModuleContainerTest
+{
+    public function createModuleContainer(array $rootDefinitions, array $moduleDefinitions): array
+    {
+        $compositeContainer = new CompositeContainer();
+        //$rootContainer = new Container($rootDefinitions, [],  $compositeContainer);
```
samdark, line 16: Remove?

yiiliveext: This is just an example of a module container. We haven't any valid implementation now, so I wrote the abstract test.

tests/Unit/AbstractModuleContainerTest.php
yiiliveext: They have many small but important differences.

samdark:
This PR will be updated soon. Problems in tests are valid and to be solved.

PR #147, opened 2020-09-10
Fix typo in tests

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/tests/Unit/ContainerTest.php
+++ b/tests/Unit/ContainerTest.php
@@ -218,7 +218,7 @@ class ContainerTest extends TestCase
     public function testCircularReference(): void
     {
         $container = new Container([
-            TreeItem::class, TreeItem::class,
+            TreeItem::class => TreeItem::class,
         ]);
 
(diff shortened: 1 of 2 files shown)


tests/Unit/ContainerTest.php
```diff
@@ -218,7 +218,7 @@ public function testDependencies(): void
     public function testCircularReference(): void
     {
         $container = new Container([
-            TreeItem::class, TreeItem::class,
+            TreeItem::class => TreeItem::class,
```
mikehaertl, line 221: @samdark Maybe there should be a check for `is_string()` in `setMultiple() to avoid these type of errors?

samdark: Yeah.

mikehaertl: Added a check.

samdark:
Thanks!

Issue #148, opened 2020-09-16
[RFC] Support factories in definition


BoShurik:
It would be great to have something like this:
```php
return [
    Api::class => [
        '!factory' => [ApiFactory::class, 'create'], // Autowiring support
    ],
];
```

```php
return [
    Api::class => function (ContainerInterface $container) {
        $foo = $container->get(Foo::class);

        return $container->get(ApiFactory::class)->create($foo);
    },
];
```

https://symfony.com/doc/current/service_container/factories.html
https://php-di.org/doc/php-definitions.html#factories

yiiliveext:
Try this

```php
class Api
{
    public function create(Foo $foo)
    {
        return new Api($foo, true, '678khjhkj'); 
    }
}
```

```php
    ApiFactory::class => [new ApiFactory(), 'create'],
    Api::class => function (ContainerInterface $container) {
        return $container->get(ApiFactory::class);
   },
];
```

BoShurik:
Yes, there are lots of workaround. I suggest another one that quite common in another DI-containers.
Also it can get rid of closures and simplify serialisation

hiqsol:
I like the idea!
It could be something like this:

```php
return [
    Api::class => Reference::toCall(ApiFactory::class, 'create'), // with autowiring support
];
```

Also it can support different ways to specify the call:

```php
/// with factory given as ready object
Reference::toCall(new ApiFactory(), 'create');

/// with factory given as previously defined id
Reference::toCall('api-factory', 'create');

/// with provided method arguments
Reference::toCall(ApiFactory::class, 'create', ['option' => 'value']); /// need to think how to merge given arguments with autowired ones
```

And it can be easily implemented with extending `Yiisoft\Factory\Definitions\CallableDefinition`

hiqsol:
@BoShurik 
Injection in closures was added and now it is simply:

```php
return [
    Api::class => fn (ApiFactory $factory) => $factory->create(),
    /// with arguments too
    Api::class => fn (ApiFactory $factory) => $factory->create('option'),
    /// static too
    Api::class => fn () => ApiFactory::staticCreate('option'),
    /// also callable as array was available before
    Api::class => [ApiFactory::class, 'staticCreate'],
];
```

I think it is better then the proposed array syntax.
Please, close the issue if it is solved.

samdark:
I agree with @hiqsol.

BoShurik:
[...]

PR #150, opened 2020-09-18
Allow more complex function call based defitions

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #148 

Add use of injector to allow more complex definitions.
See https://github.com/yiisoft/factory/pull/35 for examples.

Also add `ContainerInterface` definition in container by default
(but it still can be redefined).

Also extracted out `ReferenceTest` from huge `ContainerTest` which can be split up even more later.

This PR requires https://github.com/yiisoft/factory/pull/35 for tests to pass.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -54,6 +54,9 @@ final class Container extends AbstractContainerConfigurator implements Container
         ContainerInterface $rootContainer = null
     ) {
         $this->setMultiple($definitions);
+        if (!$this->has(ContainerInterface::class)) {
+            $this->set(ContainerInterface::class, $rootContainer ?? $this);
+        }
         $this->addProviders($providers);
         if ($rootContainer !== null) {
             $this->delegateLookup($rootContainer);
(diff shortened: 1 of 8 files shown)


src/Container.php
```diff
@@ -54,12 +55,26 @@ public function __construct(
...
+        }
         $this->addProviders($providers);
         if ($rootContainer !== null) {
             $this->delegateLookup($rootContainer);
         }
     }
 
+    private $injector;
+
+    public function getInjector()
```
yiiliveext, line 69: This is an interface violation

xepozz (approved):

samdark:
Need to run benchmark on master and this PR to ensure performance is not affected. @hiqsol would you please do it?

samdark:
Benchmark results are alright. No significant changes.

PR #152, opened 2020-09-23
Add alternative tags implementation

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

This PR needs https://github.com/yiisoft/factory/pull/47 for tests to pass.

### DIFF
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -21,12 +21,27 @@ final class CompositeContainer implements ContainerInterface
 
     public function get($id)
     {
+        if (strncmp($id, '#', 1) === 0) {
+            return $this->collectTags($id);
+        }
         foreach ($this->containers as $container) {
             if ($container->has($id)) {
                 return $container->get($id);
             }
         }
         throw new NotFoundException("No definition for $id");
+    }
+
+    private function collectTags(string $id): array
+    {
+        $res = [];
+        foreach ($this->containers as $container) {
+            if ($container->has($id)) {
+                $res = array_merge($res, $container->get($id));
+            }
+        }
+
+        return $res;
     }
 
     public function has($id): bool
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -34,14 +34,14 @@ final class CompositeContainer implements ContainerInterface
 
     private function collectTags(string $id): array
     {
-        $res = [];
+        $results = [];
         foreach ($this->containers as $container) {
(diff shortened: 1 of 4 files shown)


BoShurik:
In this implementation you must manually collect all tagged services. See https://github.com/yiisoft/di/issues/125#issuecomment-645310275

BoShurik:
BTW it works out of the box:
```php
$container = new Container([
    'engines' => [
        '__class' => ArrayIterator::class,
        '__construct()' => [
            [Reference::to(Foo::class), Reference::to(Bar::class)],
            0
        ],
    ],
]);

var_dump($container->get('engines'));
```

hiqsol:
@BoShurik the PR was refactored, added support for same syntax and tests as original implementation

src/CompositeContainer.php
```diff
@@ -29,6 +32,18 @@ public function get($id)
...
+    private function collectTags(string $id): array
+    {
+        $res = [];
+        foreach ($this->containers as $container) {
+            if ($container->has($id)) {
+                $res = array_merge($res, $container->get($id));
+            }
+        }
+
+        return $res;
```
samdark, line 44:
```suggestion
        $results = [];
        foreach ($this->containers as $container) {
            if ($container->has($id)) {
                $results = array_merge($results, $container->get($id));
            }
        }

        return $results;
```

tests/Unit/TagsTest.php
```diff
@@ -0,0 +1,214 @@
...
+use Yiisoft\Factory\Exceptions\InvalidConfigException;
+
+/**
+ * ContainerTest contains tests for \Yiisoft\Di\Container
+ */
+class TagsTest extends TestCase
+{
+    public function testTagNameAlreadyInUse(): void
+    {
+        $this->expectException(InvalidConfigException::class);
```
samdark, line 26: Why?

hiqsol: Cause `engine` is already taken for `EngineMarkOne`
Missed hash sign could be a typo.

It is another question should we allow names without hash for tags.
This implementation mostly doesn't care about # in tag names. 
The hash sign is only needed for `CompositeContainer`.

samdark: We should not allow non-hashed names, I think.

samdark:
Closing in favor of https://github.com/yiisoft/di/pull/144

PR #153, opened 2020-09-24
Add getInstances and setInstances AbstractContainerConfigurator

mj4444ru:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | 

This additional functionality allows for non-standard applications to interfere with the work of **Yiisoft\Di\Container**. Now there is a **set** method that allows you to interfere with the work of **Yiisoft\Di\Container**, so this new functionality does not threaten anything bad.

```php
class TestClass1
{
    public function __construct()
    {
        echo "\n*** Create TestClass1 ***\n\n";
    }
}

class TestClass2
{
    public function __construct()
    {
        echo "\n*** Create TestClass2 ***\n\n";
    }
}

class TestContainerConfigurator extends Yiisoft\Di\AbstractContainerConfigurator
{
    private array $ids;

    public function __construct(array $ids)
    {
        $this->ids = $ids;
    }

    public function reset(Container $container): void
    {
        $container->setInstances(\array_fill_keys($this->ids, null), true);
    }
}

$config = [
    TestClass1::class => TestClass1::class,
    TestClass2::class => TestClass2::class,
    TestContainerConfigurator::class => [
        '__construct()' => [[
            TestClass2::class,
        ]],
    ],
];
$providers = [];
$container = new Container($config, $providers);

var_dump($container->get(TestClass1::class));
var_dump($container->get(TestClass2::class));
echo "\n=== Next Request ===\n\n";
$container->get(TestContainerConfigurator::class)->reset($container);
var_dump($container->get(TestClass1::class));
var_dump($container->get(TestClass2::class));
```

We get the following output:
```
*** Create TestClass1 ***

object(TestClass1)#8 (0) {
}

*** Create TestClass2 ***

object(TestClass2)#9 (0) {
}

=== Next Request ===

object(TestClass1)#8 (0) {
}

*** Create TestClass2 ***

object(TestClass2)#4 (0) {
}
```

This functionality may help to avoid the introduction into a classic php web application of the code necessary for the application to work under conditions of multiple requests (such as RoadRunner), since there will be no need to clear the data after processing the request.

### DIFF
--- a/src/AbstractContainerConfigurator.php
+++ b/src/AbstractContainerConfigurator.php
@@ -22,4 +22,14 @@ abstract class AbstractContainerConfigurator
     {
         throw new \RuntimeException("Method 'delegateLookup' does not exist.");
     }
+
+    protected function getInstances(): array
+    {
+        throw new \RuntimeException("Method 'getInstances' does not exist.");
+    }
+
+    protected function setInstances(array $instances, bool $merge = false): void
+    {
+        throw new \RuntimeException("Method 'setInstances' does not exist.");
+    }
 }
--- a/src/Container.php
+++ b/src/Container.php
@@ -31,7 +31,7 @@ final class Container extends AbstractContainerConfigurator implements Container
     private array $building = [];
 
     /**
-     * @var object[]
+     * @var object[]|null[]
      */
     private array $instances = [];
 
@@ -138,7 +138,7 @@ final class Container extends AbstractContainerConfigurator implements Container
 
     /**
      * Returns the currently created instances.
-     * @return object[]
+     * @return object[]|null[]
      */
     protected function getInstances(): array
     {
--- a/src/Container.php
+++ b/src/Container.php
@@ -137,6 +137,25 @@ final class Container extends AbstractContainerConfigurator implements Container
     }
 
     /**
+     * Returns the currently created instances.
+     * @return object[]
+     */
+    protected function getInstances(): array
+    {
+        return $this->instances;
+    }
+
+    /**
+     * Replaces instances.
+     * @param object[]|null[] $instances
+     * @param bool $merge specifies to install instances or merge with existing ones.
+     */
+    protected function setInstances(array $instances, bool $merge = false): void
+    {
+        $this->instances = $merge ? \array_merge($this->instances, $instances) : $instances;
+    }
+
+    /**
      * Creates new instance by either interface name or alias.
      *
      * @param string $id The interface or an alias name that was previously registered.


yiiliveext:
Rather you should add the `reset` method to `AbstractConfigurator` or create a separate `ResetableContainer` with the public `reset` method.

mj4444ru:
> Rather you should add the `reset` method to `AbstractConfigurator` or create a separate `ResetableContainer` with the public `reset` method.

This can be applied not only for this.
The ResetableContainer implementation can be tricky and come with pitfalls.

samdark:
This idea is definitely interesting...

> This can be applied not only for this.

Where are other use cases?

yiiliveext:
> > Rather you should add the `reset` method to `AbstractConfigurator` or create a separate `ResetableContainer` with the public `reset` method.
> 
> This can be applied not only for this.

Your implementation with `setInstances` and `getInstances` is a very bad idea. Instances management must be encapsulated inside `Container` class. No one should know about instances inside the container. External instances management leads to container corruption.

> The ResetableContainer implementation can be tricky and come with pitfalls.

I think, you just don't understand `ResetableContainer` principle. When you call `reset` method of `ResetableCOntainer`, the container just looks over the instances and if an instance implements ResetableInterface then `reset` method of this instance is called.

samdark:
> I think, you just don't understand ResetableContainer principle. When you call reset method of ResetableCOntainer, the container just looks over the instances and if an instance implements ResetableInterface then reset method of this instance is called.

That would require proxy-classes, right? Because we can't require user to implement a certain interface. Especially if it's vendor code.

mj4444ru:
> Your implementation with `setInstances` and `getInstances` is a bad idea. Instances management must be encapsulated inside `Container` class. No one should know about instances inside the container. External instances management leads to container corruption.

If we want to damage the container, we can pass a closure to it through the set method. Thus, this functionality does not pose any danger, but it provides additional opportunities.

> I think, you just don't understand `ResetableContainer` principle. When you call `reset` method of `ResetableCOntainer`, the container just looks over the instances and if an instance implements `ResetableInterface` then `reset` method of this instance is called.

For example, I want to write a microservice using third-party code, I do not want to adapt it for yii3, which may not be possible at all because of the final classes. I just want to delete some objects without any hassle.

Perhaps I want to cache the objects I initialized and install cloned copies for each separate cycle. I can write my own container for this, but we can make life easier for people without understanding the security level of the framework.

mj4444ru:
> This idea is definitely interesting...
> 
> > This can be applied not only for this.
> 
> What are other use cases?

You can think of many different options. To summarize them, here are some of them.
- Fast removal of some of the previously created objects from the container.
- Pre-initialization of objects and the use of their cloned copies in a loop.
- Obtaining a list of created container objects for debugging purposes.

When developing a microservice, we can focus on developing the microservice itself, and not on adding classes to the architecture of the framework to implement resettable states.

samdark:
Closing since resetter is now implemented.

PR #159, opened 2020-09-26
Switch to psalm and fix static analysis issues

hiqsol:
There is one issue in `Container` still.

`$this->rootContainer->attach($container)`

It requires more substantial changes.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/CompositeContainer.php
```diff
@@ -19,13 +19,10 @@ final class CompositeContainer implements ContainerInterface
      */
     private array $containers = [];
 
-    public function get($id, array $parameters = [])
+    public function get($id)
```
samdark, line 22: Was it redundant?

hiqsol: yes, we've removed params from get everywhere

PR #161, opened 2020-09-26
Add circular exception case 1

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

This test case should be fail


hiqsol:
#162 substitutes this test and this PR should be closed?

PR #166, opened 2020-09-28
Small refactor, fix tests

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


xepozz:
Test case is here https://github.com/yiisoft/di/pull/167

samdark:
@xepozz would you please resolve conflicts?

hiqsol:
Taken part of this refactoring in https://github.com/yiisoft/di/commit/4042b039f024a384633dac018be7587cead02dd5

src/Container.php
```diff
@@ -51,16 +51,21 @@ public function __construct(
         array $definitions = [],
         array $providers = [],
         ContainerInterface $rootContainer = null
-    ) {
+    )
+    {
         $this->delegateLookup($rootContainer);
+
+        $this->set(ContainerInterface::class, $this->rootContainer ?? $this);
```
samdark, line 58: Was `has()` check unnecessary? Root container may have it defined.

src/Container.php
```diff
@@ -51,16 +51,21 @@ public function __construct(
...
+        $this->set(Injector::class, new Injector($this->rootContainer ?? $this));
```
samdark, line 59: Same here. Root container may have it defined.

src/Container.php
```diff
@@ -51,16 +51,21 @@ public function __construct(
...
-        if (!$this->has(ContainerInterface::class)) {
-            $this->set(ContainerInterface::class, $rootContainer ?? $this);
-        }
-        $this->addProviders($providers);
 
+        $this->addProviders($providers);
         # Prevent circular reference to ContainerInterface
-        $this->get(ContainerInterface::class);
+        $container = $this->get(ContainerInterface::class);
+        if ($container !== $this) {
```
samdark, line 66: Can't we simply check for `$this->rootContainer !== null`?

samdark (changes requested):
Seems it does not pass tests.

src/Container.php
xepozz: > Root container may have it defined

I don't understand you. Can you please explain what do you mean?

src/Container.php
xepozz: It's not just a check.
It's combine the fix (early fetching container `$this->get(ContainerInterface::class);`) and redefining `Injector` if container definition was modified.

src/Container.php
samdark: You have removed a `has()` call. I think it was there to check if there is a definition in a root container already:

```php
$rootContainer = new MyRootContainer();
$rootContainer->define(ContainerInterface::class, function () {
    return new SuperContainer();
});
$yiiContainer = new Container($defintions, $providers, $rootContainer);
```

I am not entirely sure what was the check for. Therefore asking.

samdark:
The idea was mostly included into other PRs so closing. @xepozz thanks for direction.

PR #173, opened 2020-10-01
Add tests for constructor parameters passed properly

hiqsol:
- excessive parameters must be dropped  
- if parameter is variadic all values must be passed  
- mixing integer and string indexed parameters is not allowed  

| Q             | A
| ------------- | ---
| Is bugfix?    | ?
| New feature?  | ❌
| Breaks BC?    | ❌


samdark:
From a user perspective I'd prefer an exception in case I'm specifying non-existing parameter rather than passing it to constructor.

samdark (approved):

PR #178, opened 2020-10-06
Add default definitions

hiqsol:
Improve code style.
Make default definitions explicit and collected in single place.
Allow change `Injector` used in the container (with test).

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -53,14 +53,20 @@ final class Container extends AbstractContainerConfigurator implements Container
         ContainerInterface $rootContainer = null
     ) {
         $this->delegateLookup($rootContainer);
+        $this->setDefaultDefinitions();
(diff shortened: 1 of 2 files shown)


samdark (changes requested):
It breaks yii-bootstrap4 tests with:

```
Error: Class 'Yiisoft\Injector\Injector' not found

C:\src\yii-dev-tool\dev\di\src\Container.php:68
C:\src\yii-dev-tool\dev\di\src\Container.php:56
C:\src\yii-dev-tool\dev\yii-bootstrap4\tests\TestCase.php:43
```

samdark (approved):
Re-verified. Works.

PR #189, opened 2020-11-11
Fixed bug with $building.

mj4444ru:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -15,6 +15,7 @@ use Yiisoft\Factory\Exceptions\NotFoundException;
 use Yiisoft\Factory\Exceptions\NotInstantiableException;
 use Yiisoft\Injector\Injector;
 
+use function array_key_exists;
 use function array_keys;
 use function assert;
 use function class_exists;
@@ -98,7 +99,11 @@ final class Container extends AbstractContainerConfigurator implements Container
      */
     public function get($id)
     {
-        return $this->instances[$id] ?? ($this->instances[$id] = $this->build($id));
+        if (!array_key_exists($id, $this->instances)) {
+            $this->instances[$id] = $this->build($id);
+        }
+
+        return $this->instances[$id];
     }
 
     /**
--- a/src/Container.php
+++ b/src/Container.php
@@ -15,6 +15,12 @@ use Yiisoft\Factory\Exceptions\NotFoundException;
 use Yiisoft\Factory\Exceptions\NotInstantiableException;
 use Yiisoft\Injector\Injector;
 
+use function array_keys;
+use function assert;
+use function class_exists;
+use function implode;
+use function is_string;
+
 /**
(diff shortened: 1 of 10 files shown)


src/Container.php
```diff
@@ -157,8 +154,11 @@ private function build(string $id)
         }
 
         $this->building[$id] = 1;
-        $object = $this->buildInternal($id);
-        unset($this->building[$id]);
+        try {
+            $object = $this->buildInternal($id);
```
mj4444ru, line 158: When an error occurred, `$this->building` was not cleared and an error occurred on re-access.

src/Container.php
```diff
@@ -157,8 +154,11 @@ private function build(string $id)
         }
 
         $this->building[$id] = 1;
-        $object = $this->buildInternal($id);
-        unset($this->building[$id]);
```
mj4444ru, line 161: When an error occurred, `$this->building` was not cleared and an error occurred on re-access.

src/Container.php
```diff
@@ -157,8 +154,11 @@ private function build(string $id)
...
+        try {
+            $object = $this->buildInternal($id);
+        } finally {
+            unset($this->building[$id]);
+        }
```
mj4444ru, line 161: When an error occurred, `$this->building` was not cleared and an error occurred on re-access.

samdark:
Need a unit test that tests for the behavior changed.

tests/Unit/ContainerTest.php
```diff
@@ -872,6 +872,33 @@ public function register(Container $container): void
         );
     }
 
+    public function testBuildingVar(): void
+    {
+        $building = null;
+        $container = new Container([
+            ColorPink::class => static function (ContainerInterface $container) use (&$building): ColorPink {
+                $building = ((array)$container)["\000Yiisoft\Di\Container\000building"];
```
samdark, line 880: Is it possible to test it via public API only without accessing internal state?

tests/Unit/ContainerTest.php
```diff
@@ -872,6 +872,33 @@ public function register(Container $container): void
...
+        ]);
+        $container->get(ColorPink::class);
+        $this->assertSame([ColorPink::class => 1], $building);
+    }
+
+    public function testBuildingVarClean(): void
+    {
+        $container = new Container();
+        $container->get(ColorPink::class);
+        $this->assertCount(0, ((array)$container)["\000Yiisoft\Di\Container\000building"]);
```
samdark, line 892: Is it possible to test it via public API only without accessing internal state?

samdark (approved):
The fix is logically correct but tests are accessing internal state which isn't good at all. I wonder if you had an issue with the public API or was it spotted during reading code?

samdark:
Thank you! Great catch.

PR #202, opened 2021-02-01
Refine README and fix some grammar

mikehaertl:
Revised README

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ and configure classes resolving dependencies.
 - Supports aliasing.
 - Supports service providers and deferred service providers.
 
+
 ## Using the container
 
 Usage of the DI container is fairly simple: You first initialize it with an
@@ -96,8 +97,9 @@ better to rely on autowiring as provided by the Injector available from the
 
 ## Using aliases
 
-The DI container supports aliases via the `Reference` class. This way objects
-can also be retrieved by a more handy name:
+The DI container supports aliases via the
+`Yiisoft\Factory\Definitions\Reference` class. This way objects can also be
+retrieved by a more handy name:
 
 ```php
 use Yiisoft\Di\Container;
@@ -111,9 +113,12 @@ $object = $container->get('engine_one');
 
 ## Delegated lookups and composite containers
 
-The `Container` class supports delegated lookups. This means, that dependencies
-are always fetched from a given *root container*. It can be passed as third
-parameter to the container constructor:
+Another feature of the `Container` class are *delegated lookups*. This means
+that *all* dependencies for definitions in the container should be resolved via
+a *root container* - and not by the container itself.
+
+To use delegated lookups a root container can be passed as third argument to
+the constructor:
 
 ```php
 class Car
@@ -130,14 +135,25 @@ class Car
         return $this->engine;
     }
 }
-$rootContainer = new Container([
-    EngineInterface::class => EngineMarkOne::class
-]);
-$container = new Container([], [], $rootContainer);
 
+$rootContainer = new Container([
+    EngineInterface::class => EngineMarkTwo::class
+]);
+$container = new Container([
+    EngineInterface::class => EngineMarkOne::class,
+], [], $rootContainer);
+
+// returns an instance of `Car`
 $car = $container->get(Car::class);
-$engine = $car->getEngine(); //returns an instance of the `Car` class
+// returns an instance of `EngineMarkTwo`
+$engine = $car->getEngine();
 ```
+
+Note, that the root container is only used for resolving dependencies. You can
+not directly fetch entries of the root container from the container via `get()`.
+
+Delegated lookups are mainly useful for composite containers.
+
 
 ### Composite containers
 
@@ -353,7 +369,7 @@ application if you perform heavy operations inside the `register()` method.
 
 ## Using deferred service providers
 
-To prevent the potential perfomance decrease when using service providers you
+To prevent the potential performance decrease when using service providers you
 can use so-called *deferred service providers*.
 
 They extend from `Yiisoft\Di\Support\DeferredServiceProvider` and must
@@ -430,13 +446,11 @@ the container will first check the result of the `provides()` method.
 Because `EngineInterface` is listed there it will then call the `register()`
 method of the `CarFactoryProvider`.
 
-**Note**, you can use deferred service providers not just to defer bootstrap of
-heavy services but also to register your services to the container only when
-they are actually needed.
 
 ## Further reading
 
 - [Martin Fowler's article](http://martinfowler.com/articles/injection.html).
+
 
 # Benchmarks
 
--- a/README.md
+++ b/README.md
@@ -210,37 +210,6 @@ $car = $composite->get(CarInterface::class);
 $engine = $composite->get(EngineInterface::class);
 ```
 
-A composite container can be a *root container* for delegate container lookups.
-
-```php
-use Yiisoft\Di\CompositeContainer;
-use Yiisoft\Di\Container;
-
-$container = new CompositeContainer();
-$container1 = new Container([
-    'first' => static function () {
-        return 'first';
-    },
-    'third' => static function () {
-        return 'third';
-    }
-]);
-$container2 = new Container([
-    'second' => static function () {
-        return 'second';
-    },
-    'first-and-second-and-third' => static function ($c) {
-        return $c->get('first') . ' ' . $c->get('second') . ' ' . $c->get('third');
-    },
-]);
-
-$container->attach($container1);
-$container->attach($container2);
-$first = $container->get('first'); // returns 'first'
-$second = $container->get('second'); // returns 'second'
-$firstSecondThird = $container->get('first-and-second-and-third'); // returns 'first second third' 
-```
-
 ## Contextual containers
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -63,35 +68,36 @@ return [
...
+Note, however, that it is bad practice to use a container directly. It's much
+better to rely on autowiring as provided by the Injector available from the
+[yiisoft/injector](https://github.com/yiisoft/injector) package.
+
 
 ## Using aliases
 
-Container supports aliases via `Reference` class. It could be useful to have an ability
-to retrieve objects both by their interface and named explicitly:
+The DI container supports aliases via the `Reference` class. This way objects
```
mikehaertl, line 99: @samdark It's unclear what is meant by `Reference` class here. I couldn't find this class in the repo. Can you explain?

README.md
```diff
@@ -100,24 +106,25 @@ $container = new Container([
...
 
-## Delegated lookup and composite containers
+## Delegated lookups and composite containers
 
-The `Container` class supports delegated lookup.
-When using delegated lookup, all dependencies are always fetched from a given root container.
-To use delegate lookup you should set `root container` as a third parameter of the container constructor.
+The `Container` class supports delegated lookups. This means, that dependencies
+are always fetched from a given *root container*. It can be passed as third
+parameter to the container constructor:
```
mikehaertl, line 116: IMO the example following here is not really helpful to understand what delegate lookups are used for (at least not for me). 

```
$rootContainer = new Container([
    EngineInterface::class => EngineMarkOne::class
]);
$container = new Container([], [], $rootContainer);

$car = $container->get(Car::class);
$engine = $car->getEngine(); //returns an instance of the `Car` class
```

What's the point of `$container` here? Why not use `$rootContainer` directly?

README.md
```diff
@@ -166,18 +178,24 @@ $carContainer = new Container([
...
-$car = $composite->get(CarInterface::class); //returns an instance of a `Car` class
-$engine = $composite->get(EngineInterface::class); //returns an instance of a `EngineMarkTwo` class
+// returns an instance of a `Car` class
+$car = $composite->get(CarInterface::class);
+// returns an instance of a `EngineMarkTwo` class
+$engine = $composite->get(EngineInterface::class);
 ```
 
-A composite container can be a `root container` for a container delegate lookup.
+A composite container can be a *root container* for delegate container lookups.
```
mikehaertl, line 197: This is also a bit unclear and the example doesn't really help. The way "root container" was explained above I was expecting here, that a composite container is passed as 3rd argument to a new container somehow.

The example just fetches data from the composite container as shown before: 

```
    'first-and-second-and-third' => static function ($c) {
        return $c->get('first') . ' ' . $c->get('second') . ' ' . $c->get('third');
    },
```

I assume that `$c` is the composite container here? This could be made more clear I think. Still I'm missing how this is related to "root container"

README.md
```diff
@@ -395,13 +423,16 @@ $engine = $container->get(EngineInterface::class);
...
+`EngineInterface` gets requested from the container. When this happens,
+the container will first check the result of the `provides()` method.
+Because `EngineInterface` is listed there it will then call the `register()`
 method of the `CarFactoryProvider`.
 
-**Note**, you can use deferred service providers not just to defer bootstrap of heavy services but also to register your 
-services to the container only when they are actually needed. 
+**Note**, you can use deferred service providers not just to defer bootstrap of
+heavy services but also to register your services to the container only when
+they are actually needed.
```
mikehaertl, line 435: What type of "your service" is meant here? If it's any "heavy service" then this is what was just explained (i.e. tautology).

README.md
samdark: It's this one: https://github.com/yiisoft/factory/blob/master/src/Definitions/Reference.php

README.md
samdark: It works like that:

```php
$rootContainer = new Container([
    EngineInterface::class => EngineMarkTwo::class,
]);
$container = new Container([
], [], $rootContainer);

$car = $container->get(Car::class);
$engine = $car->getEngine(); //returns an instance of `EngineMarkOne`
```
i.e. what's specified in `get()` is being looked for in the main container but dependencies of what's found are taken from container we delegate to.

The feature use-case is described here: https://github.com/container-interop/container-interop/blob/master/docs/Delegate-lookup-meta.md#42-typical-usage

README.md
samdark: See https://github.com/yiisoft/di/pull/202#discussion_r568180393

README.md
samdark: Yes, that's repeating... let's remove it.

README.md
mikehaertl: Thanks, that link really helped a lot. So I would expect the following logic:
1. Dependency is defined in both, `$rootContainer` and `$container`: Use entry from `$rootContainer`
2. Dependency is only in `$rootContainer`: Use entry from `$rootContainer`
3. Dependency is only in `$container` but *not* in `$rootContainer`: Use entry from `$container`

But it seems 3. is not the case: There's no fallback to `$container`. Why is that and wouldn't this make sense?

README.md
mikehaertl: Still not clear, sorry. From the article:

> The delegate container will usually be a composite container. A composite container is a container that contains several other containers. When performing a lookup on a composite container, the inner containers are queried until one container returns an entry. 

The critical part to me is this:
> An inner container implementing the delegate lookup feature will return entries it contains, but if these entries have dependencies, the dependencies lookup calls will be performed on the composite container, giving a chance to all containers to answer.

I'm missing this part in the implementation. I would expect that when you attach an inner container to the composite, the inner container will somehow get the composite itself as root container. That's not the case I think. So an inner container will only use it's own definitions and never query the composite container that it's attached to.

Is this correct or did I miss it somehow?

README.md
mikehaertl: Added full class name

README.md
mikehaertl: Updated, but my question remains open.

I think there was a little mistake in your example: It's EngineMarkTwo, as defined in the root container.

README.md
mikehaertl: Removed

src/Contracts/ServiceProviderInterface.php
```diff
@@ -21,23 +21,22 @@
  * {
  *    public function register(Container $container): void
  *    {
- *        $this->registerDependencies();
- *        $this->registerService();
+ *        $this->registerDependencies($container);
+ *        $this->registerService($container);
```
mikehaertl, line 25: @samdark Also fixed this here, please check.

samdark: All good 👍

README.md
samdark: Yes, right. I've made a mistake in the comment. Should be `EngineMarkTwo`.

samdark: The logic is:

1. When getting a service, get it directly from non-root container.
2. When getting service sub-dependency, try get it from root container first. Then, if there was nothing found, check in non-root container.

See also https://github.com/yiisoft/di/blob/master/tests/Unit/ContainerTest.php#L921.

README.md
samdark:
```php
$containerOne = new Container([EngineInterface::class => EngineMarkOne::class]);
$containerTwo = new Container([EngineInterface::class => EngineMarkTwo::class]);
$compositeContainer = new CompositeContainer();
$compositeContainer->attach($containerOne);
$compositeContainer->attach($containerTwo);

$car = $compositeContainer->get(Car::class);
$this->assertInstanceOf(EngineMarkTwo::class, $car->getEngine());
```

Test passes.

README.md
mikehaertl: I think point 2) is not the case. I had checked the test before. If you remove `EngineInterface::class` definition from the `rootContainer` the test will fail:

```php
    public function testContainerDelegateLookupContainer(): void
    {
        $rootContainer = new Container(
            [
               // EngineInterface::class => EngineMarkTwo::class,
            ]
        );

        $container = new Container(
            [
                'injected' => static fn (Injector $injector) => $injector->make(Car::class),
            ],
            [],
            $rootContainer
        );
        $car = $container->get(Car::class);
        $c = $container->get(ContainerInterface::class);
        $injected = $container->get('injected');

        $this->assertSame(Car::class, get_class($car));
        $this->assertSame(EngineMarkTwo::class, get_class($car->getEngine()));
        $this->assertSame(EngineMarkTwo::class, get_class($injected->getEngine()));
        $this->assertSame(CompositeContainer::class, get_class($c));
    }
```

```
There was 1 error:

1) Yiisoft\Di\Tests\Unit\ContainerTest::testContainerDelegateLookupContainer
Yiisoft\Factory\Exceptions\NotFoundException: No definition for Yiisoft\Di\Tests\Support\EngineInterface
```

README.md
mikehaertl: That's not my point. My problem is this sentence in the current README:

> A composite container can be a root container for a container delegate lookup.
> [example here]

I can not see how the example relates to "root delegation". It only demonstrates composite containers. Where does root delegation happpen with composite containers?

samdark:
Thanks! I've removed controversial part of it.

PR #204, opened 2021-03-13
Psalm: fix errors + improve annotations for Container::get()

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

- Fixed psalm errors.
- Added annotations for `Contaner::get()` which allow the psalm to understand what the method returns.

For example:

```php
$handler = $container->get(ErrorHandler::class);
```

Psalm will be understand that `$handler` is instance of `ErrorHandler`.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -105,11 +105,11 @@ final class Container extends AbstractContainerConfigurator implements Container
      * @throws NotFoundException
      * @throws NotInstantiableException
      *
-     * @return object An instance of the requested interface.
+     * @return mixed|object An instance of the requested interface.
      *
      * @psalm-template T
      * @psalm-param string|class-string<T> $id
-     * @psalm-return ($id is class-string ? T : object)
+     * @psalm-return ($id is class-string ? T : mixed)
      */
     public function get($id)
     {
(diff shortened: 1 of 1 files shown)


src/Container.php
```diff
@@ -106,6 +106,10 @@ public function has($id): bool
      * @throws NotInstantiableException
      *
      * @return object An instance of the requested interface.
+     *
+     * @psalm-template T
+     * @psalm-param string|class-string<T> $id
+     * @psalm-return ($id is class-string ? T : object)
```
roxblnfk, line 112:
```suggestion
     * @psalm-return ($id is class-string ? T : mixed)
```

roxblnfk (commented):
I think it is more correct

src/Container.php
vjik: Are you sure that this implementation of container by can return not object?

vjik: Done.

samdark (approved):

samdark:
👍

PR #227, opened 2021-07-29
Use AbstractContainerConfigurator instead of Container

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | 

This allows to use another implementation of Yiisoft Container or any bridges to connect Yiisoft Container with another containers


samdark:
@xepozz waiting till @yiiliveext will finish providers re-work.

PR #235, opened 2021-08-05
Move injector to dependency resolver

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -

This is adopt PR to https://github.com/yiisoft/factory/pull/123 and instead of https://github.com/yiisoft/di/pull/222


src/DependencyResolver.php
```diff
@@ -56,8 +58,21 @@ public function resolve(string $id)
...
     {
         return false;
     }
+
+    private function getInjector(): Injector
+    {
+        if ($this->injector === null) {
+            $this->injector = new Injector($this->container);
+        }
+        return $this->injector;
```
xepozz, line 76:
```suggestion
        return $this->injector ??= new Injector($this->container);
```

src/DependencyResolver.php
```diff
@@ -56,8 +58,21 @@ public function resolve(string $id)
...
+    {
+        return $this->getInjector()->invoke($callable);
+    }
+
     public function shouldCloneOnResolve(): bool
     {
         return false;
     }
+
+    private function getInjector(): Injector
```
xepozz, line 71: What about to move the initialization into __construct method?

src/DependencyResolver.php
vjik: Done.

src/DependencyResolver.php
vjik: injector may not be needed

samdark (approved):

PR #244, opened 2021-08-17
Replace string literals with class names

Nex-Otaku:
String literal keys in this code sample looks like definition of properties instead of definition of class dependencies.

Replacing it with class names should help understanding this sample.

| Q             | A
| ------------- | ---
| Is bugfix?    | No
| New feature?  | No
| Breaks BC?    | No
| Fixed issues  | -

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,16 +56,16 @@ The configuration can be stored in a `.php` file that returns an array:
 ```php
 return [
     EngineInterface::class => EngineMarkOne::class,
-    'full_definition' => [
+    FullDefinitionObject::class => [
         'class' => EngineMarkOne::class,
         '__construct()' => [42], 
         '$propertyName' => 'value',
         'setX()' => [42],
     ],
-    'closure' => fn (SomeFactory $factory) => $factory->create('args'),
-    'static_call_preferred' => fn () => MyFactory::create('args'),
-    'static_call_supported' => [MyFactory::class, 'create'],
-    'object' => new MyClass(),
+    ClosureObject::class => fn (SomeFactory $factory) => $factory->create('args'),
+    StaticCallPrefferedObject::class => fn () => MyFactory::create('args'),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,16 +56,16 @@ The configuration can be stored in a `.php` file that returns an array:
 ```php
 return [
-    'full_definition' => [
+    FullDefinitionObject::class => [
```
BoShurik, line 59:
```diff
- FullDefinitionObject::class => [
+ EngineMarkOne::class => [
```

README.md
```diff
@@ -56,16 +56,16 @@ The configuration can be stored in a `.php` file that returns an array:
...
         'setX()' => [42],
     ],
-    'closure' => fn (SomeFactory $factory) => $factory->create('args'),
-    'static_call_preferred' => fn () => MyFactory::create('args'),
-    'static_call_supported' => [MyFactory::class, 'create'],
-    'object' => new MyClass(),
+    ClosureObject::class => fn (SomeFactory $factory) => $factory->create('args'),
+    StaticCallPrefferedObject::class => fn () => MyFactory::create('args'),
+    StaticCallSupportedObject::class => [MyFactory::class, 'create'],
+    CustomObject::class => new MyClass(),
```
BoShurik, line 68:
```diff
- CustomObject::class => new MyClass(),
+ MyClass::class => new MyClass(),
```

BoShurik (changes requested):

BoShurik:
`FullDefinitionObject` and `CustomObject` must be subclass of `EngineMarkOne` and `MyClass` respectively . It's not obvious and may be confusing

xepozz:
> String literal keys in this code sample looks like definition of properties instead of definition of class dependencies.

Can you please describe why? I think it doesn't. 

Your version doesn't show that user can use not only full class name as a key of definition.

samdark:
I think @xepozz is right. There are enough other places through the guide to show how to use `::class`.

Issue #228, opened 2021-07-31
Incorrect exception on trying to get object with undefined exception


vjik:
Test https://github.com/yiisoft/di/blob/2cc9f803ece970bbbb5d8ccf740c3312e7580d1b/tests/Unit/ContainerTest.php#L570 :

```php
$container = new Container(
	[
		'car' => Car::class,
	]
);

$this->expectException(NotFoundException::class);
$container->get('car');
```

Should throws `NotInstantiableException` instead of `NotFoundException`.

samdark:
Seems to contradict PSR:

![photo_2021-11-20_00-56-08](https://user-images.githubusercontent.com/47294/142698064-d843f1e5-4943-44a6-81ea-be00dbf6435a.jpg)

PR #264, opened 2021-10-13
Fix incorrect exception on trying to get object with undefined exception

dehbka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/di/issues/228


samdark:
See issue.

PR #274, opened 2021-11-23
Improve CompositeContainer error

yiiliveext:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/CompositeContainer.php
+++ b/src/CompositeContainer.php
@@ -63,6 +63,10 @@ final class CompositeContainer implements ContainerInterface
             }
         }
 
+        if ($this->containers !== []) {
+            $this->containers[0]->get($id);
+        }
+
         throw new NotFoundException($id);
     }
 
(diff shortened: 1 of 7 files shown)


src/CompositeContainer.php
```diff
@@ -63,6 +63,10 @@ public function get($id)
             }
         }
 
+        if ($this->containers !== []) {
```
samdark, line 66: May do a foreach here and throw CompositeNotFoundException:

1. Yiisoft\Di\Container #22 : error message
2. Yiisoft\Cycle\RepositoryContainer #34: error message
...
5. Some\Another\Container #77: error message

src/CompositeNotFoundException.php
```diff
@@ -15,6 +15,7 @@ final class CompositeNotFoundException extends Exception implements NotFoundExce
 {
     /**
      * @param Throwable[] $exceptions Exceptions of parent containers.
+     * @psalm-param array $exceptions
```
vjik, line 20: `@param Throwable[] $exceptions` should be enough

PR #293, opened 2021-12-20
Fix class methods when used extensible service

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -8,11 +8,11 @@ use Closure;
 use InvalidArgumentException;
 use Psr\Container\ContainerInterface;
 use Yiisoft\Definitions\ArrayDefinition;
+use Yiisoft\Definitions\DefinitionStorage;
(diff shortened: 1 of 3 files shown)


src/Container.php
```diff
@@ -584,7 +584,9 @@ private function addProviders(array $providers): void
                 /** @var mixed $definition */
                 $definition = $this->definitions->get($id);
                 if (!$definition instanceof ExtensibleService) {
-                    $definition = new ExtensibleService($definition);
+                    $definition = new ExtensibleService(
```
vjik, line 587: I think better for performance pass `$definition` and `$id` to `ExtensibleService` without normalization and into `ExtensibaleService` use `Yiisoft\Di\Helpers\DefinitionNormalizer` instead of `Yiisoft\Definitions\Helpers\Normalizer`.

Issue #312, opened 2022-11-05
Check tag services array in `Container::setTags()` on empty


vjik:
Now we can add tag with empty array and `$container->has()` returns true. It's wrong.

xepozz:
I think it's totally okay to have an empty tags.
First of all, we shouldn't put sticks in the wheels. Making a developer remove a line or start a debugging research do it.
An empty tag may be computed during some operations from config (removing keys or values).
There are nothing bad if sometime tag can be empty. It's ok to have it. You may provide a point of the expansion of your library. 
Also tagged definitions are `iterable`. Empty tag is also iterable.

vjik:
> I think it's totally okay to have an empty tags. First of all, we shouldn't put sticks in the wheels. Making a developer remove a line or start a debugging research do it. An empty tag may be computed during some operations from config (removing keys or values). There are nothing bad if sometime tag can be empty. It's ok to have it. You may provide a point of the expansion of your library. Also tagged definitions are `iterable`. Empty tag is also iterable.

Should  `$container->has()` returns `true` on empty tag?

xepozz:
For sure

vjik:
Ok, agree

PR #313, opened 2022-11-05
Fix incorrect result of `Container::has()` on check empty tag

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 ## 1.0.3 June 17, 2022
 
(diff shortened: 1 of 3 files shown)


xepozz (changes requested):
I don't like it, more comments in the issue

vjik:
> I don't like it, more comments in the issue

Agree, fixed.

PR #315, opened 2022-11-09
Add test for wrong definition format

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Depends on | https://github.com/yiisoft/definitions/pull/54


vjik:
Now depends on https://github.com/yiisoft/definitions/pull/57

PR #316, opened 2022-11-09
Fix exception messages

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Extends | https://github.com/yiisoft/di/pull/315
| Depends on | https://github.com/yiisoft/definitions/pull/56

### DIFF
--- a/tests/Unit/ContainerTest.php
+++ b/tests/Unit/ContainerTest.php
@@ -381,7 +381,7 @@ final class ContainerTest extends TestCase
 
         $this->expectException(BuildingException::class);
         $this->expectExceptionMessage(
-            'An error "Arguments indexed both by name and by position are not allowed in the same array." occurred while building "test".'
+            'Caught unhandled error "Arguments indexed both by name and by position are not allowed in the same array." while building "test".'
         );
         $container->get('test');
     }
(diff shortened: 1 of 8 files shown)


tests/Unit/ContainerTest.php
```diff
@@ -1580,7 +1584,10 @@ public function getExtensions(): array
             }
         };
 
-        $this->expectException(RuntimeException::class);
+        $this->expectException(BuildingException::class);
+        $this->expectExceptionMessage(
+            'Caught unhandled error "" while building "Yiisoft\Di\Tests\Support\B".'
```
samdark, line 1589: Empty error?

xepozz: RuntimeException doesn't provide any messages

samdark: Why do we need `""` then in the message?

samdark (approved):

samdark:
Tests seems to fail: https://github.com/yiisoft/di/actions/runs/3604857565/jobs/6074657047

PR #317, opened 2022-11-30
Fix delegated container

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️

### DIFF
--- a/tests/Unit/ContainerTest.php
+++ b/tests/Unit/ContainerTest.php
@@ -1235,6 +1235,34 @@ final class ContainerTest extends TestCase
         $this->assertSame(42, $engine->getNumber());
     }
 
+    public function testNewContainerDefinitionInDelegates(): void
+    {
+        $firstContainer = null;
+        $secondContainer = null;
+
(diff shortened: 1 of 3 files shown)


samdark:
Need a test for it.

tests/Unit/ContainerTest.php
```diff
@@ -1235,6 +1235,34 @@ static function (ContainerInterface $container) {
...
     }
 
+    public function testNewContainerDefinitionInDelegates(): void
+    {
+        $firstContainer = null;
+        $secondContainer = null;
+
+        $config = ContainerConfig::create()
+            ->withDefinitions([
+                ContainerInterface::class => new Container(ContainerConfig::create()),
```
rustamwin, line 1245: What is it for?

xepozz: For the test case :)
If you change the `ContainerInterface` definition, it should be used everywhere.

samdark (approved):

vjik (approved):

Issue #333, opened 2023-07-19
Proposal to Add Reference::tag Method


nahakyuu:
I propose adding a new method, `Reference::tag`

```php
// before
Reference::to('tag@translation.categorySource');
Reference::to('tag@' . Yiisoft\Log\Target::class);
// after
Reference::tag('translation.categorySource');
Reference::tag(Yiisoft\Log\Target::class);
```

vjik:
Can we use one of exist class for implement suggested method or create new?

BoShurik:
Maybe add a `tag` function which will use `Reference::to('tag@'. $value);`?

PR #353, opened 2024-02-13
Add shortcut for tag reference #333

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #333


src/Helpers/TagHelper.php
```diff
@@ -1,21 +0,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace Yiisoft\Di\Helpers;
-
-/**
- * @internal
- */
-final class TagHelper
```
vjik, line 10: Why need to make methods `extractTagFromAlias` and `isTagAlias` public? May be keep this internal helper?

xepozz: Because `tag@` is now backed inside the di repo. It's easy to rename it any time if it's needed.
Users may build their own logic on top of the tags and it's good to provide them an api to check and verify tags.

tests/Unit/Reference/TagReferenceTest.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->assertFalse(TagReference::isTagAlias('tag#test'));
+        $this->assertTrue(TagReference::isTagAlias('tag@test'));
+    }
+
+    public function testExtractTag(): void
+    {
+        $this->assertEquals('test', TagReference::extractTagFromAlias('tag@test'));
+    }
+
+    public function testExtractWrongTag(): void
```
vjik, line 26: Second expressions don't check. Need split to two tests or use data provider.

vjik (changes requested):
Need to add line to changelog and info about `TagReference` to readme.

samdark (approved):

samdark:
👍

PR #354, opened 2024-02-13
Add ability to use container as a factory

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    |/❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #350

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -35,7 +35,8 @@ final class Container implements ContainerInterface
 {
     private const META_TAGS = 'tags';
     private const META_RESET = 'reset';
-    private const ALLOWED_META = [self::META_TAGS, self::META_RESET];
+    private const META_AFTER_BUILT = 'afterBuilt';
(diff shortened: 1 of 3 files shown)


src/Container.php
```diff
@@ -35,7 +35,8 @@ final class Container implements ContainerInterface
 {
     private const META_TAGS = 'tags';
     private const META_RESET = 'reset';
-    private const ALLOWED_META = [self::META_TAGS, self::META_RESET];
+    private const META_AFTER_BUILT = 'afterBuilt';
```
vjik, line 38: Are there any other cases for this hook?

xepozz: Not now

vjik (commented):
If we know only one case for new hook (use container as factory for defined classes), I suggest don't add hook now. For this case more convient and clear will be use specific meta tag `singleton` with boolean value (`true` as default).

```php
'singleton' => false
```

It is better than:

```php
'afterBuilt' => AfterBuiltHook::unsetInstance(),
```

Issue #324, opened 2023-02-11
Make `BuildingException` friendly


vjik:
And if nested exception is friendly — show solution.

Issue #234, opened 2021-08-05
Use friendly exceptions


vjik:
See https://github.com/yiisoft/friendly-exception

PR #386, opened 2025-01-12
Use friendly exception

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #324

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - New #380: Add `TagReference::id()` method (@vjik)
 - Enh #384: Make `$config` parameter in `Container` constructor optional (@np25071984)
+- Enh #324: Make `BuildingException` friendly (@np25071984)
 
 ## 1.3.0 October 14, 2024
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - New #380: Add `TagReference::id()` method (@vjik)
 - Enh #384: Make `$config` parameter in `Container` constructor optional (@np25071984)
-- Enh #324: Make `BuildingException` friendly (@np25071984)
+- Enh #324: Make `BuildingException` and `NotFoundException` friendly (@np25071984)
 
 ## 1.3.0 October 14, 2024
 
--- a/src/BuildingException.php
+++ b/src/BuildingException.php
@@ -36,7 +36,7 @@ final class BuildingException extends Exception implements ContainerExceptionInt
 
     public function getName(): string
     {
-        return 'Couldn\'t build requested object.';
+        return 'Unable to build object requested.';
     }
 
     public function getSolution(): ?string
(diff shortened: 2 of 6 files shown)


np25071984:
Not sure if https://github.com/yiisoft/di/issues/234 and https://github.com/yiisoft/di/issues/324 are equal

src/BuildingException.php
```diff
@@ -32,4 +33,16 @@ public function __construct(
 
         parent::__construct($message, 0, $previous);
     }
+
+    public function getName(): string
+    {
+        return 'Couldn\'t build requested object.';
```
np25071984, line 39: I was having a hard time coming up with decent texts here. Definitely help is needed...

src/BuildingException.php
```diff
@@ -32,4 +33,16 @@ public function __construct(
...
+    }
+
+    public function getSolution(): ?string
+    {
+        return <<<SOLUTION
+            See (https://github.com/yiisoft/di)[https://github.com/yiisoft/di] for more documentation.
```
samdark, line 45: Let's think about what could cause it:

1. Class wasn't found.
2. Config wasn't valid.
3. What else?

I'd provide a list of things to try, such as:

1. Ensure that either a service with ID "x" is defined or such class exists and is autoloadable.
2. Ensure that configuration for service with ID "x" is correct.

Docs are not suggesting on how to fix errors.

vjik (changes requested):
`BuildingException` throws rarely. In most cases will be throwed `NotFoundException`. It needs to be made friendly.

src/NotFoundException.php
```diff
@@ -49,4 +50,20 @@ public function getBuildStack(): array
     {
         return $this->buildStack;
     }
+
+    public function getName(): string
+    {
+        return sprintf('No difinition or class found for "%s" ID.', $this->id);
```
samdark, line 56:
```suggestion
        return sprintf('No definition or class found for "%s" ID.', $this->id);
```

src/NotFoundException.php
```diff
@@ -49,4 +50,20 @@ public function getBuildStack(): array
...
+    }
+
+    public function getSolution(): ?string
+    {
+        $solution = <<<SOLUTION
+            Ensure that either a service with ID "%1\$s" is defined or such class exists and is autoloadable.
+
+            Ensure that configuration for service with ID "%1\$s" is correct.
```
samdark, line 64: Is this exception thrown for incorrect configurations?

samdark:
:+1:

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - New #380: Add `TagReference::id()` method (@vjik)
 - Enh #384: Make `$config` parameter in `Container` constructor optional (@np25071984)
+- Enh #324: Make `BuildingException` friendly (@np25071984)
```
samdark, line 7: Add `NotFoundException` please.

samdark:
Looks good to me now. @vjik ?

vjik (approved):
I would also like to display build stack beautifully, but we can make it in future.

samdark:
Thank you!

PR #387, opened 2025-02-09
Performance optimization

samdark:
Before:

```
\Yiisoft\Di\Tests\Benchmark\ContainerMethodHasBench

    benchPredefinedExisting.................R1 I2 - Mo132.604μs (±1.14%)
    benchUndefinedExisting..................R1 I2 - Mo124.260μs (±0.57%)
    benchUndefinedNonexistent...............R1 I4 - Mo455.709μs (±0.86%)

\Yiisoft\Di\Tests\Benchmark\ContainerBench

    benchConstruct..........................R1 I4 - Mo320.728μs (±1.86%)
    benchSequentialLookups # 0..............R3 I3 - Mo620.365μs (±0.44%)
    benchSequentialLookups # 1..............R3 I4 - Mo643.744μs (±0.72%)
    benchSequentialLookups # 2..............R1 I0 - Mo644.136μs (±1.16%)
    benchRandomLookups # 0..................R1 I4 - Mo623.952μs (±0.69%)
    benchRandomLookups # 1..................R1 I4 - Mo649.820μs (±1.61%)
    benchRandomLookups # 2..................R2 I4 - Mo643.426μs (±1.05%)
    benchRandomLookupsComposite # 0.........R2 I4 - Mo2.315ms (±0.71%)
    benchRandomLookupsComposite # 1.........R2 I4 - Mo2.330ms (±1.03%)
    benchRandomLookupsComposite # 2.........R1 I4 - Mo2.317ms (±1.05%)
```

After:

```
\Yiisoft\Di\Tests\Benchmark\ContainerMethodHasBench

    benchPredefinedExisting.................R5 I4 - Mo88.505μs (±1.04%)
    benchUndefinedExisting..................R2 I2 - Mo86.247μs (±1.94%)
    benchUndefinedNonexistent...............R2 I4 - Mo454.755μs (±1.42%)

\Yiisoft\Di\Tests\Benchmark\ContainerBench

    benchConstruct..........................R2 I4 - Mo219.828μs (±2.02%)
    benchSequentialLookups # 0..............R1 I0 - Mo494.586μs (±0.64%)
    benchSequentialLookups # 1..............R1 I4 - Mo541.868μs (±0.67%)
    benchSequentialLookups # 2..............R1 I4 - Mo540.820μs (±2.01%)
    benchRandomLookups # 0..................R1 I1 - Mo520.341μs (±0.51%)
    benchRandomLookups # 1..................R1 I4 - Mo541.799μs (±1.23%)
    benchRandomLookups # 2..................R1 I0 - Mo547.495μs (±1.06%)
    benchRandomLookupsComposite # 0.........R1 I4 - Mo2.215ms (±1.22%)
    benchRandomLookupsComposite # 1.........R1 I1 - Mo2.170ms (±1.19%)
    benchRandomLookupsComposite # 2.........R1 I2 - Mo2.174ms (±0.65%)
```


src/Container.php
```diff
@@ -212,12 +239,16 @@ private function addDefinition(string $id, mixed $definition): void
         [$definition, $meta] = DefinitionParser::parse($definition);
         if ($this->validate) {
             $this->validateDefinition($definition, $id);
-            $this->validateMeta($meta);
+            // Only validate meta if it's not empty.
+            if ($meta !== []) {
```
xepozz, line 243: not necessary as for me, foreach perfectly works with empty arrays

src/Container.php
```diff
@@ -136,65 +149,79 @@ public function has(string $id): bool
...
-                                array_unshift($buildStack, $id);
-                                throw new NotFoundException($exception->getId(), $buildStack);
-                            }
-                            throw $exception;
-                        }
-                        throw new NotFoundException($id, [$id], previous: $exception);
-                    }
+        // Fast path: check if instance exists.
+        if (isset($this->instances[$id])) {
+            return $id === StateResetter::class ? $this->prepareStateResetter($id) : $this->instances[$id];
```
xepozz, line 154: Why isn't just a `return $instances[$id]`?

`prepareStateResetter` already holds values in the `$instances`, isn't it?

xepozz (changes requested):

src/Container.php
vjik: `prepareStateResetter` check delegates before.

src/Container.php
```diff
@@ -136,65 +149,79 @@ public function has(string $id): bool
...
-                }
-                $mainResetter->setResetters($resetters);
-            } elseif ($delegatesResetter !== null) {
-                $resetter = new StateResetter($this->get(ContainerInterface::class));
-                $resetter->setResetters([$mainResetter, $delegatesResetter]);
+    /**
+     * @param string $id
+     * @return mixed
+     */
+    private function prepareStateResetter(string $id)
```
vjik, line 195:
```suggestion
    private function prepareStateResetter(): StateResetter
```

`$id` is always `StateResetter::class`

src/Container.php
```diff
@@ -136,65 +149,79 @@ public function has(string $id): bool
...
-                                $buildStack = $exception->getBuildStack();
-                                array_unshift($buildStack, $id);
-                                throw new NotFoundException($exception->getId(), $buildStack);
-                            }
-                            throw $exception;
-                        }
-                        throw new NotFoundException($id, [$id], previous: $exception);
-                    }
+        // Fast path: check if instance exists.
+        if (isset($this->instances[$id])) {
```
vjik, line 153: Use `array_key_exists` instead of `isset`. Theoretically, result may be `null`.

src/Container.php
```diff
@@ -136,65 +149,79 @@ public function has(string $id): bool
...
+            throw new NotFoundException($exception->getId(), $buildStack);
+        } catch (NotFoundExceptionInterface $exception) {
+            // Try delegates before giving up
+            return $this->delegates->has($id) ? $this->delegates->get($id) : throw new NotFoundException($id, [$id], previous: $exception);
+        } catch (ContainerExceptionInterface $e) {
+            if (!$e instanceof InvalidConfigException) {
+                throw $e;
             }
+            throw new BuildingException($id, $e, $this->definitions->getBuildStack(), $e);
+        } catch (Throwable $e) {
```
vjik, line 178: Seems we should catch any exceptions from previous "catch" blocks also (for example, exceptions from delegates).

src/Container.php
vjik: Call a method has cost.

src/Container.php
```diff
@@ -295,26 +327,24 @@ private function setDelegates(array $delegates): void
...
-            $methodsAndProperties = $definition['methodsAndProperties'];
-
-            $definition = array_merge(
-                $class === null ? [] : [ArrayDefinition::CLASS_NAME => $class],
-                [ArrayDefinition::CONSTRUCTOR => $constructorArguments],
-                // extract only value from parsed definition method
-                array_map(static fn (array $data): mixed => $data[2], $methodsAndProperties),
-            );
+        // Skip validation for common simple cases.
+        if (is_string($definition) || $definition instanceof ContainerInterface || $definition instanceof Closure) {
```
vjik, line 331: We should validate string definitions (on empty).

Validation process is not required hard performance optimization because he is not running in production environment.

src/Container.php
```diff
@@ -492,15 +519,28 @@ private function build(string $id)
...
+        if (!$this->definitions->has($id)) {
+            throw new NotFoundException($id, $this->definitions->getBuildStack());
+        }
+
         $this->building[$id] = 1;
         try {
-            if (!$this->definitions->has($id)) {
-                throw new NotFoundException($id, $this->definitions->getBuildStack());
+            // Use cached normalized definition if available.
+            if (!isset($this->normalizedDefinitions[$id])) {
```
vjik, line 535: What case when needs to normalize definition not once?

vjik (changes requested):

src/Container.php
xepozz: So why wouldn't you inline the whole framework in a single function?

src/Container.php
xepozz: All delegates should be calculated and set one time, why they should do it every call?

vjik: > All delegates should be calculated and set one time, why they should do it every call?

I means check delegated containters. See current [code](https://github.com/yiisoft/di/blob/911dbb01486f52f2193f185682caf0f65549744d/src/Container.php#L167):

```php
if ($id === StateResetter::class) {
    $delegatesResetter = null;
    if ($this->delegates->has(StateResetter::class)) {
        $delegatesResetter = $this->delegates->get(StateResetter::class);
    }
```

src/Container.php
samdark: It makes sense since there are many definitions and cost of calling a method is higher than cost of the check.

samdark:
Did some fixes. @vjik can you help with Psalm? I can ignore mixed returns but, I guess, it could be solved in a better way.

vjik:
> Did some fixes. @vjik can you help with Psalm? I can ignore mixed returns but, I guess, it could be solved in a better way.

Fixed. I also suppresed `MixedReturnStatement` because mixed return type is OK for `get()` method.

src/Container.php
samdark: Do you have a test for it in mind?

vjik: See https://github.com/yiisoft/di/pull/388

samdark:
@vjik test merged and is now passing with latest commit.

src/CompositeContainer.php
```diff
@@ -103,11 +103,43 @@ public function get($id)
...
+            );
+        }
+
+        if ($id === StateResetter::class) {
+            foreach ($this->containers as $container) {
+                if ($container->has(StateResetter::class)) {
+                    return true;
+                }
+            }
+            return false;
```
vjik, line 122:
```suggestion
            return true;
```

src/Container.php
```diff
@@ -103,15 +108,23 @@ public function __construct(?ContainerConfigInterface $config = null)
...
         if (TagReference::isTagAlias($id)) {
             $tag = TagReference::extractTagFromAlias($id);
             return isset($this->tags[$tag]);
         }
 
         try {
-            return $this->definitions->has($id);
-        } catch (CircularReferenceException) {
-            return true;
+            return $this->delegates->has($id);
```
vjik, line 125: I suggest to add this check in another PR. It is required more tests.

src/Container.php
```diff
@@ -295,26 +342,26 @@ private function setDelegates(array $delegates): void
...
+            return;
         }
 
-        if ($definition instanceof ExtensibleService) {
+        if (is_array($definition)) {
+            if (isset($definition[DefinitionParser::IS_PREPARED_ARRAY_DEFINITION_DATA])) {
+                $class = $definition['class'];
+                $constructorArguments = $definition['__construct()'];
+
+                /** @psalm-var array<string,mixed> $methodsAndProperties */
```
vjik, line 355: keep comments

src/Container.php
```diff
@@ -295,26 +342,26 @@ private function setDelegates(array $delegates): void
...
+                $methodsAndProperties = $definition['methodsAndProperties'];
+
+                $definition = array_merge(
+                    $class === null ? [] : [ArrayDefinition::CLASS_NAME => $class],
+                    [ArrayDefinition::CONSTRUCTOR => $constructorArguments],
+                    array_map(static fn (array $data): mixed => $data[2], $methodsAndProperties),
```
vjik, line 361: keep comments

vjik (changes requested):

src/Container.php
samdark: Do you mean returning `false` at this stage instead of checking delegates?

vjik: Yes. There no this feature before.

samdark: Done.

samdark:
All comments solved. @vjik anything else?

vjik:
> All comments solved. @vjik anything else?

Looks good. It remains to add the tests. Test coverage is reduced.

vjik (approved):

